PR #43, opened 2022-04-12
DB-1868: Add pnpm+changesets GitHub action

CobyPear:
- Installed changesets bot and configured for this repo https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md#recommended-automation-flow
- Setup pnpm chagesets recipe https://pnpm.io/using-changesets
- Updated changelog

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+name: Changesets
+on:
+  push:
+    branches:
+      - main
+      - canary
+env:
+  CI: true
+  PNPM_CACHE_FOLDER: .pnpm-store
+jobs:
+  version:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: setup node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+
+      - name: install pnpm
+        run: npm i pnpm@latest -g
+
+      - name: Setup npmrc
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
+name: Changesets
+on:
+  push:
+    branches:
+      - main
+      - canary
```
backlineint, line 6: If we're only publishing based on main, should canary be removed from the list here?

CHANGELOG.md
```diff
@@ -5,9 +5,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+## [1.0.0] - 2022-04-11
```
backlineint, line 8: This is currently curated manually, correct? If we keep up with using changesets, I wonder if a future enhancement could be to populate this automatically based on the changes since the last publish? Or at the very least, having individual changelogs in each of the sub-projects should make it easier to populate this manually without forgetting things.

backlineint (changes requested):

.github/workflows/release.yml
CobyPear: True, good catch!

CHANGELOG.md
CobyPear: Yes, there is a way to automate the changelog with chagesets - there are some plugin type things we could use or roll our own, but I recall seeing one that will source changes via GitHub.

Something else we should discuss: Should we enforce changesets on all PRs?

backlineint: @CobyPear I think we should try enforcing changesets for PRs. If we don't, I worry that we won't keep up with them.

PR #48, opened 2022-04-13
Add lock files to starter kits for distribution

backlineint:
Added lock files to each of the starters. Tested to confirm that it didn't make pnpm mad.

Ideally we wouldn't have to do this, but in the meantime it will at least unblock things. We'll either need to remember to update the locks manually before we merge to main, or automate something if this becomes a long term fix.


backlineint:
That changeset detection is nice :)

CobyPear (approved):

PR #53, opened 2022-04-13
DB-1868: Add publish script to package.json

CobyPear:
- Adds publish script to root package.json for use with changeset action
- Updates changeset action to v1 from master

I misunderstood this part of the docs but it's more clear now 🙃 

> By adding publish: pnpm ci:publish which is a script that executes changeset publish will publish to the registry once the PR is opened by changeset version.

https://pnpm.io/using-changesets

I'm still not confident this is correct; Do we need a `ci:publish` script like so?

```json
"publish:ci": "pnpm changeset version"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "build:all": "pnpm recursive run build",
     "generate-docs": "pnpm --filter ./web clear && pnpm --filter ./web build",
     "postinstall": "pnpm build:pkgs",
-    "publish": "pnpm ci:publish"
+    "ci:version": "changeset version",
+    "ci:publish": "changeset publish"
   },
   "repository": {
     "type": "git",


package.json
```diff
@@ -10,7 +10,8 @@
     "build:starters": "pnpm --filter ./starters run build",
     "build:all": "pnpm recursive run build",
     "generate-docs": "pnpm --filter ./web clear && pnpm --filter ./web build",
-    "postinstall": "pnpm build:pkgs"
+    "postinstall": "pnpm build:pkgs",
```
backlineint, line 13: @CobyPear yeah, the docs for this seem unnecessarily confusing. If I'm parsing things correctly, the workflow is going to run the scripts for:

```
pnpm ci:version
pnpm ci:publish
```

and the scripts should be:

```
"ci:version": "changeset version",
"ci:publish": "changeset publish",
```

I technically haven't run `changeset publish` before. I used `pnpm publish -r`, but the docs seem to imply that both should work.

backlineint (changes requested):

backlineint (approved):

PR #63, opened 2022-04-15
Update IMAGE_DOMAIN Variable in .example.env

backlineint:
The name of this variable changed, so this needs to be updated too, correct?


CobyPear (approved):
Ah yes, did not mean to change this but I'm ok with it if you are.

PR #68, opened 2022-04-15
Build valid lock files for starters

backlineint:


CobyPear (approved):
The lockfile version is 2 but the container is looking for v1. It works with v2 but I wonder if it's any faster with v1? Looks good other than that, nice catch on the create-next-app command.

PR #72, opened 2022-04-18
DB-2114: [next-drupal] Fix .env.local loading in production

CobyPear:
According to [Next.js env variable load order](https://nextjs.org/docs/basic-features/environment-variables#environment-variable-load-order), `.env.local` is loaded in production environments.  To avoid this, I've changed the local .env file to `.env.development.local` which will not be loaded in production environments.


backlineint (approved):
@CobyPear merging this so it can be tested, but I think this will also require some updates to documentation. Could you follow up with that at some point?

PR #80, opened 2022-04-19
[next-drupal] Expose Next FRONTEND_URL

CobyPear:
Previously this was done in the .env file but it makes sense to move it to the config. Currently setting FRONTEND_URL won't have the intended effect, but this change should make it work.

- Strip trailing slash from FRONTEND_URL and set it to NEXT_PUBLIC_FRONTEND_URL
- Uses NEXT_PUBLIC_FRONTEND_URL for setting the hrefLang to avoid double slashes

### DIFF
--- a/.changeset/fast-rules-occur.md
+++ b/.changeset/fast-rules-occur.md
@@ -1,0 +1,5 @@
+---
+"@pantheon-systems/next-drupal-starter": patch
+---
+
+Expose FRONTEND_URL to properly set hrefLang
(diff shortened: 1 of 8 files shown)


backlineint (changes requested):
@CobyPear with this change if the FRONTEND_URL envar is not set, Next will fail with the following error:

error - Failed to load next.config.js, see more info here https://nextjs.org/docs/messages/next-config-error
TypeError: Cannot read properties of undefined (reading 'replace')
    at Object.<anonymous> (/Users/brianperry/repos/decoupled-kit-js/starters/next-drupal-starter/next.config.js:28:65)

backlineint (approved):

PR #105, opened 2022-05-20
[next-drupal] DB-1570: Nextjs preview endpoint redirects based on content not slug

CobyPear:
## What changes were made?
The `preview` endpoint now verifies content exists and redirect based on the content's path instead of the slug from the query param.

A custom 500 page was added to redirect to in case an error occurs while verifying preview data.

I was trying to utilize our global stores to set the error and display it on the error page, however I was having some problems getting the state in the error page, I think I'm missing something on how those work.

Also clearing params on the page list page to avoid errors for example when the data is prefetched while on the homepage.

next-drupal-starter

Tested preview against both decoupled profiles (the QA CMS sites) with the starter running locally

## Additional information


starters/next-drupal-starter/pages/500.js
```diff
@@ -0,0 +1,13 @@
+import Link from "next/link";
+
+export default function Custom500() {
```
backlineint, line 3: I think this is an improvement for now, but what this really feels like is a preview error route rather than a generic 500.

I think I'm going to create an enhancement ticket for this. If this was a preview error route, we could do things like pass params to provide feedback like 'the path does not match' or 'the requested preview path does not exist'

starters/next-drupal-starter/pages/api/preview.js
```diff
@@ -1,34 +1,54 @@
...
+  // returns the store that matches the locale found in the requested url
+  // or the only store if using a monolingual backend
+  const [store] = globalDrupalStateAuthStores.filter(({ defaultLocale }) => {
+    const regex = new RegExp(`/${defaultLocale}/`);
+    return defaultLocale ? regex.test(req.url) : true;
+  });
+
+  // get the object name from the slug
+  const regex = new RegExp(
+    `^(?:/${store.defaultLocale}/|/)(?<objectName>.*)/.*$`
```
backlineint, line 19: This works for our out of the box setup, but I worry that it will be too restrictive as people build actual sites. For example, this wouldn't support a path like `/about` or any case where the first path argument doesn't have a 1:1 relation with the objectName. 

To fix this, my best guess is that we'd need to update the module to pass the object name as an additional query parameter. Since the open redirect can be considered a vulnerability, I'm going to merge the current state of this, and create tickets for future refactoring.

starters/next-drupal-starter/pages/500.js
backlineint: Created https://getpantheon.atlassian.net/browse/DB-2874 for this.

starters/next-drupal-starter/pages/api/preview.js
backlineint: Created:
https://getpantheon.atlassian.net/browse/DB-2875
https://getpantheon.atlassian.net/browse/DB-2876

backlineint (approved):

PR #106, opened 2022-05-24
[next-drupal] DB-2339: Abstract page templates into components

CobyPear:
## What changes were made?
DRY up some code by creating reusable components.
Some components still have some major assumptions about the data coming in, but I'm not sure how to address that at the moment.

next-drupal-starter

Tested locally. Paginator tested with recipes data and no routing.

## Additional information

It would be cool to make the paginator a [HOC](https://reactjs.org/docs/higher-order-components.html) that takes in a component that would be used to render the data, but I couldn't figure it out in a reasonable amount of time.


starters/next-drupal-starter/components/article.js
```diff
@@ -0,0 +1,34 @@
+import { IMAGE_URL } from "../lib/constants";
+import Image from "next/image";
+import Link from "next/link";
+
+export default function Article({ article }) {
```
backlineint, line 5: I see what you're saying about assumptions about the shape of the data. This example is pretty simple so it isn't the end of the world, but it still illustrates the problem.

In the past I've addressed this by making components like this as presentational as possible and accepting the exact props they need rather than a CMS specific object. In this example the props would be: imgSrc, title, body. That makes this an article as in `<article>` not `node--article`. That would allow someone to use this article component even if their field was named something like `field_hero_image` In fact, it is so generic that this component be used with another CMS and/or abstracted into something like the 'react-kit` package we've considered previously.

Something still needs to handle the logic to prepare the data for this presentational component. I've used a provider-style component for that in the past. Simplified example:

```js
export default function ArticleProvider({ node }) {
  const imgSrc = article?.field_media_image?.field_media_image?.uri?.url || "";
  // handle the other props here
    return (
       <Article imgSrc={imgSrc} />
    )
}
```

Maybe that would be better expressed as a hook these days.

```js
const {imgSrc, title, body} = useArticle(node);

<Article
  imgSrc={imgSrc}
  title={title}
  etc...
/>
```

Details to work out for sure especially as we start thinking about the provider or hook, but that is the basic idea. Thoughts?

starters/next-drupal-starter/components/grid-list.js
```diff
@@ -0,0 +1,66 @@
+import { IMAGE_URL } from "../lib/constants";
+
+import Grid from "./grid";
+import Image from "next/image";
+import Link from "next/link";
+
+export default function GridList({ contentArr, contentType, multiLanguage }) {
```
backlineint, line 7: What was described for Article probably even applies more here. The `field_recipe_category` is really just a category (or tag) and it should be possible to use this grid with other category-like fields.

starters/next-drupal-starter/components/paginator.js
```diff
@@ -0,0 +1,212 @@
+import { useState, useEffect } from "react";
```
backlineint, line 1: This makes so much more sense as a component :)

starters/next-drupal-starter/pages/articles/[...slug].js
```diff
@@ -1,50 +1,23 @@
...
 import { isMultiLanguage } from "../../lib/isMultiLanguage";
 import { fetchJsonapiEndpoint } from "@pantheon-systems/drupal-kit";
 import {
   getCurrentLocaleStore,
   globalDrupalStateAuthStores,
 } from "../../lib/drupalStateContext";
 
-import Image from "next/image";
-import Link from "next/link";
+import Article from "../../components/article.js";
```
Comment (backlineint) [inline starters/next-drupal-starter/pages/articles/[...slug].js:9]: Really excited about how clean these templates will look after this refactoring, and also some of the stuff we're planning on doing related to static paths and preview.

backlineint (changes requested):
Left some comments inline about how we might abstract things a bit more to make these components less dependent on the shape of the CMS data. Consider that me thinking out loud at this point / the start of a discussion.

This is still a big improvement, even if we agree that we want to do more to make these components re-usable. I'd be OK with merging the current state if the revisions seem like something that should be a future ticket.

One exception though - I'm not seeing the footer menu anymore. I don't think that regression was introduced here, but I'm not positive. Could you take a quick look at that?

CobyPear:
> I'm not seeing the footer menu anymore

@backlineint  I noticed this when we moved the menu fetching to the component. Looks like the backends CORS won't let `localhost:3000` through so we can't fetch the menu data from the component. This is partly what led to that 'layout hack' in the first place.

Part of the issue stems from `DrupalState.getObject` always trying to fetch `dsApiIndex` even if it already exists in state. So if we fetch the menu data on the homepage, I can see it in the store in the menu component, however the fetch to `dsApiIndex` fails and the menu Data cannot be retrieved.

When sidestepping that issue by trying to access the data directly with `.getState()['itemName']`,  I am unable to access the data in useEffect and set the component state accordingly.

I'll continue poking at it.

image

starters/next-drupal-starter/components/article.js
CobyPear: Ah yeah that makes more sense! I like the idea of a hook or HOC. Might pull the grid out into an HOC also.
Will also be more explicit with the props that are coming back from the `get*Props` functions.

backlineint:
> Part of the issue stems from `DrupalState.getObject` always trying to fetch `dsApiIndex` even if it already exists in state. So if we fetch the menu data on the homepage, I can see it in the store in the menu component, however the fetch to `dsApiIndex` fails and the menu Data cannot be retrieved.

If this is the current behavior, it isn't what I'd expect. I'd expect that we only fetch the index from Drupal the first time, like we do for objects. Could you create an issue in the d.o issue queue for this?

backlineint:
> @backlineint I noticed this when we moved the menu fetching to the component. Looks like the backends CORS won't let `localhost:3000` through so we can't fetch the menu data from the component. This is partly what led to that 'layout hack' in the first place.

Strange that would fail only for menu and not everywhere. Regardless, we do have a couple of CORS related tickets on the backlog:

https://getpantheon.atlassian.net/browse/DB-2045
https://getpantheon.atlassian.net/browse/DB-1892

I'll look at prioritizing the first one. Since that is low effort, it might be worth doing to make our lives easier.

CobyPear:
> Strange that would fail only for menu and not everywhere.

This makes sense to me because all other fetches are happening in `get*Props`, not from the client side.

starters/next-drupal-starter/components/grid-list.js
CobyPear: I tried going for a HOC here similar to the new paginator that is BYOC (bring your own component). Open to tabling this component for later if it's still not up to snuff.

starters/next-drupal-starter/pages/pages/[...alias].js
```diff
@@ -11,7 +11,7 @@ import Layout from "../../components/layout";
 const drupalUrl = process.env.backendUrl;
 
 // TODO - Much of this is duplicated in the article/[id].js file. Abstract this out into modules and components.
-export default function Home({ page, hrefLang }) {
+export default function PageTemplate({ page, hrefLang }) {
```
Comment (backlineint) [inline starters/next-drupal-starter/pages/pages/[...alias].js:14]: Thanks for cleaning these up along the way.

backlineint (changes requested):

backlineint (approved):

PR #117, opened 2022-06-01
DB-2818: Fix script path and permissions

CobyPear:
## What changes were made?
Fix the update-lockfiles permissions and path so it will run properly in GitHub Actions 🤞 

decoupled-kit

## Additional information

### DIFF
--- a/.github/workflows/release-and-split.yml
+++ b/.github/workflows/release-and-split.yml
@@ -32,10 +32,12 @@ jobs:
         run: pnpm config set store-dir $PNPM_CACHE_FOLDER
       - name: install dependencies
         run: pnpm install
+      - name: make script runnable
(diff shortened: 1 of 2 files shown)


backlineint (approved):

.github/workflows/release-and-split.yml
```diff
@@ -32,10 +32,12 @@ jobs:
         run: pnpm config set store-dir $PNPM_CACHE_FOLDER
       - name: install dependencies
         run: pnpm install
+      - name: make script runnable
+      - run: chmod +x ".github/scripts/update-lockfiles.sh"
```
backlineint, line 36: Best guess is that the leading - is the problem. I think this should be:

`run: chmod +x ".github/scripts/update-lockfiles.sh"` (without the - )

PR #129, opened 2022-06-03
DB-2295: Create helper for getStaticPaths

CobyPear:
## What changes were made?
- Added a helper function to DRY up getStaticPaths
- Implemented the helper
- Small change to these routes to use the authStore if preview is true, otherwise use the non auth'd store. This should allow builds to succeed even if credentials are incorrect.

next-drupal-starter

tested builds locally

## Additional information

### DIFF
--- a/starters/next-drupal-starter/lib/getPaths.js
+++ b/starters/next-drupal-starter/lib/getPaths.js
@@ -40,15 +40,15 @@ export const getPaths = async (
 
       // map over the data fetch to extract the path name
       return data.map((datum) => {
-        const regex = new RegExp(`^\/${urlAliasPrefix}\/(.*)$`);
-        const match = datum.path.alias.match(regex);
-        let path;
-        if (match[1]) {
-          path = match[1];
-        }
-
+        // remove the url prefix and split the path to handle
+        // dynamic routes like /articles/my-article and /articles/featured/my-article
+        const regex = new RegExp(`/?${urlAliasPrefix}/?`);
+        const path = datum.path.alias?.replace(regex, "").split("/");
         // return the path object.
-        return { params: { [`${dynamicRouteName}`]: [path] }, locale: locale };
+        return {
+          params: { [`${dynamicRouteName}`]: path },
+          locale: locale,
+        };
       });
     } catch (error) {
       // if set to failGracefully, return null.
--- a/starters/next-drupal-starter/lib/getPaths.js
+++ b/starters/next-drupal-starter/lib/getPaths.js
@@ -1,0 +1,75 @@
+import { getCurrentLocaleStore } from "./drupalStateContext";
+/**
+ * @description Helper function to get the current path nedded for getStaticPaths
+ * @param {import('next').GetStaticPathsContext} context - Nextjs getStaticPaths context
+ * @param {import('@pantheon-systems/drupal-kit').DrupalState} globalDrupalStateStores - Drupal state stores from drupalStateContext.js. Can be auth'd stores or non auth'd stores.
+ * @param {string} node - The node to fetch from Drupal. This will be passed to DrupalState.getObject as the objectName.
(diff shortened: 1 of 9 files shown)


starters/next-drupal-starter/lib/getPaths.js
```diff
@@ -0,0 +1,75 @@
...
+              path {
+                alias
+              }
+            }
+          `,
+      });
+
+      // map over the data fetch to extract the path name
+      return data.map((datum) => {
+        const regex = new RegExp(`^\/${urlAliasPrefix}\/(.*)$`);
```
backlineint, line 43: Trying to think through the cases that this regex won't handle.

* If the file system route is /pages/prefix (for example `/pages/article`) I think deeper routes like `/article/featured/my-article` will still match.

* If the file system route is deeper than one subdir - /pages/prefix/subdir (for example `/pages/article/featured`) I think routes like `/article/featured/my-article` will only work if they use `article` as the prefix, and not `article/featured` because the / needs to be escaped. Maybe we should replace `/` in urlAliasPrefix with `\/`? The alternative of explaining this is comments or docs seems like it would be difficult to do in a way that makes sense.

* I also think that this wouldn't work with paths that aren't prefixed. Some variation on removing the second `\/` in the regex might work - although you might have to make other adjustments due to the fact that the match will have a leading slash I think.

starters/next-drupal-starter/pages/articles/[...slug].js
```diff
@@ -23,50 +25,33 @@ export default function ArticleTemplate({ title, body, imgSrc, hrefLang }) {
...
 
 export async function getStaticProps(context) {
   const { locales, locale } = context;
   const multiLanguage = isMultiLanguage(locales);
   const lang = context.preview ? context.previewData.previewLang : locale;
 
-  const store = getCurrentLocaleStore(lang, globalDrupalStateAuthStores);
+  const store = getCurrentLocaleStore(
+    lang,
+    context.preview ? globalDrupalStateAuthStores : globalDrupalStateStores
```
Comment (backlineint) [inline starters/next-drupal-starter/pages/articles/[...slug].js:53]: My concern here is that there will be cases where people will want to always use the authenticated store. I'm inclined to leave this for now though - it will be an improvement in most cases and is easy to change.

backlineint (changes requested):

starters/next-drupal-starter/pages/articles/[...slug].js
Comment (CobyPear) [reply starters/next-drupal-starter/pages/articles/[...slug].js:53]: I could see that. Hoping this* will all be smoothed out when we make the necessary enhancements to DrupalState.

*as in stores per language and auth/non-auth stores

starters/next-drupal-starter/lib/getPaths.js
CobyPear:
```suggestion
         // ^(?:\/?${urlAliasPrefix}\/)? - optional capture group to match leading slash and or the urlAliasPrefix
        // ((?:\/?(.*))$ - capture group that matches path name inside optinal capture group with optional leading slash
        const regex = new RegExp(`^(?:\/?${urlAliasPrefix}\/)?((?:\/?(.*))$`);
```

Thoughts on this possible regex? `^(?:\/?${urlAliasPrefix}\/)?((?:\/?(.*))$`

starters/next-drupal-starter/pages/articles/[...slug].js
Comment (backlineint) [reply starters/next-drupal-starter/pages/articles/[...slug].js:53]: Yup was also thinking the same thing.

starters/next-drupal-starter/lib/getPaths.js
backlineint: At a glance I think that could work, although the example seems to be missing a closing paren. Instances of `/` would also still have to be handled in the urlAliasPrefix.

backlineint (approved):

PR #139, opened 2022-06-10
DB-2935: Menu Data for footer menu cannot be fetched client side

backlineint:
## What changes were made?
* Adjusted Footer component to be presentational, removed useEffect/useState, and to accept menuData as prop.
* Updated Layout component to accept footerMenu data as a prop that it will pass to the Footer component.
* Updated all page templates to source this menu data in get*Props functions. Clunky and not DRY, but as Coby found there isn't a better way in current NextJS to handle this. Probably could reduce redundant code here with some effort, but not sure it would be worth doing with nested routing and other related enhancements on Next's roadmap.

next-drupal-starter

Viewed all templates during development to confirm that footer menu rendered.

## Additional information


CobyPear (approved):
Works for me until we are able to fetch data from a layout!
Thanks for the enhancements to the error messages too.

This makes me wonder if there's an opportunity for DrupalState to be able to fetch multiple nodes at once? `getObjects` or some ability to pass an array to the `objectName` option.

PR #149, opened 2022-06-23
DB-3175: Static Heading, Footer, and initial Layout Component

enZane:
## What changes were made?
- Added components with the same look and feel as Next Drupal starter
- Added layout to index page
- Added the Pages page and Post page with WIP message

- Next WordPress starter

- Manually


backlineint (changes requested):
This looks great!

Only concern here is that the changeset doesn't quite look right. What changeset related commands did you run? Looking at the Readme I can see how this would be unclear, but you should only run `pnpm changeset` From the diff it looks like you also ran `pnpm changeset version`. That will actually happen automatically in CI.

We should change the readme to explain the creating and publishing new versions happen in CI.

In the meantime, could you revert the existing changeset commit, and then run only `pnpm changeset` After that we should be good to go.

backlineint (approved):
The changesets still didn't look right here. I'd expect there to only be one added for this feature and the existing changesets for other unreleased features should remain.

I made the adjustments so this can be merged.

PR #151, opened 2022-06-28
DB-3176: Dynamic Footer Menu

enZane:
## What changes were made?
- In the WordPress-kit, we move to use graphql-request instead of Apollo graphql
- Added the client implementation in the Next-WordPress starter
- Got menu items in the pages to pass them to the footers

- next-wordpress-starter
- wordpress-kit

- Wordpress-kit automatic test changed and runned to check the new factory
- Visual test for the next-wordpress starter


CobyPear (changes requested):
Looks like a `.DS_Store` file made it into the root of the monorepo. We should probably add this to our `.gitignore`. 

The rest looks good to me!

CobyPear (approved):

backlineint (dismissed):

starters/next-wordpress-starter/lib/Menus.js
```diff
@@ -0,0 +1,35 @@
...
+                  uri
+                  label
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  `;
```
CobyPear, line 24: I think we can get even more specific with the query to avoid the flatMap
```suggestion
    query ExampleMenuQuery {
      menu(idType: NAME, id: "Example Menu") {
        menuItems {
          nodes {
            id
            label
            path
          }
        }
      }
    }
  `;
```

enZane: Good catch! Thank you

package.json
```diff
@@ -70,6 +70,7 @@
       "trim@<0.0.3": ">=0.0.3",
       "node-forge@<1.3.0": ">=1.3.0",
       "minimist@<=1.2.5": ">=1.2.6",
+      "@pantheon-systems/wordpress-kit": "workspace:^1.0.0",
       "@pantheon-systems/drupal-kit": "workspace:^1.0.0",
```
CobyPear, line 74: Let's also make this change here, since trying to run the starters in the monorepo context will fail as is.
```suggestion
      "@pantheon-systems/wordpress-kit": "workspace:*",
      "@pantheon-systems/drupal-kit": "workspace:*",
```

CobyPear (changes requested):
@enZane Just one more small update to get the monorepo working and then this will be good to merge!

CobyPear (approved):

PR #171, opened 2022-07-20
DB-1525: [next-drupal-starter] Unit tests

CobyPear:
## What changes were made?
Added unit tests using [`vitest`](https://vitest.dev)
Fetch calls are mocked with [`msw`](https://mswjs.io)

next-drupal-starter

Locally, will check that tests are running in CI.

## Additional information

This is step 1 in backfilling tests to our stater kits. I think vitest is a great replacement for Jest that allows us to potentially share a config for all of the starters. Vitest will be used for the snapshot tests as well. The `testing-library/react` plugin has been included already in preparation for the snapshot tests.


starters/next-drupal-starter/__tests__/drupalStateContext.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+      expect(stores[0].defaultLocale).toEqual("en");
+      expect(stores[1].defaultLocale).toEqual("es");
+      expect(stores.every((store) => store.auth)).toBe(true);
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe("getCurrentLocaleStore()", () => {
+    const spy = vi.spyOn(utils, "getCurrentLocaleStore");
+    let stores;
```
enZane, line 41: Why do you have this separated? Instead of 
```
const stores = utils.makeLocaleStores({ locales }) 
```

starters/next-drupal-starter/__tests__/drupalStateContext.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  describe("makeLocaleStores()", () => {
+    const spy = vi.spyOn(utils, "makeLocaleStores");
+    it("should make a store for each locale", () => {
+      const locales = ["en", "es"];
+      const stores = utils.makeLocaleStores({ locales });
+
+      expect(stores.length).toEqual(2);
+      expect(stores[0].defaultLocale).toEqual("en");
```
enZane, line 16: In my opinion, you should only test one thing for unit test

starters/next-drupal-starter/__tests__/drupalStateContext.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+      const stores = utils.makeLocaleStores({ locales });
+      expect(stores.length).toEqual(1);
+      expect(stores.defaultLocale).toBe(undefined);
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
+    it("should make an authenticated store for each locale if given auth credentials", () => {
+      const locales = ["en", "es"];
+      const stores = utils.makeLocaleStores({ locales, auth: true });
+
+      expect(stores.length).toEqual(2);
```
enZane, line 31: I guess we don't need to retest the stores.lenght because there is another couple of test that do that

starters/next-drupal-starter/__tests__/isMultiLanguage.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+const spy = vi.spyOn(utils, "isMultiLanguage");
+
+describe("isMultiLanguage()", () => {
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+  it("should return true if there are multiple locales", () => {
+    expect(utils.isMultiLanguage(["en", "es"])).toBe(true);
+    expect(spy).toHaveBeenCalledTimes(1);
```
enZane, line 12: Which is the value to do this?

CobyPear: Which part, the spy?

enZane: Yes, and the number of called times

starters/next-drupal-starter/__tests__/drupalStateContext.test.jsx
CobyPear: Ah I think I set this up to possibly add more tests and never did. Will clean this up.

starters/next-drupal-starter/__tests__/drupalStateContext.test.jsx
CobyPear: So would you recommend 1 test for length, and another to check default locale for each store, etc?

starters/next-drupal-starter/__tests__/isMultiLanguage.test.js
CobyPear: To make sure the code we want called was called. Maybe it doesn't have a lot of value here.

CobyPear: I removed the spies. I think if we want to increase coverage we can bring them back to spy on `console.logs`, but I don't think I was using them very effectively as it was :)

starters/next-drupal-starter/__tests__/drupalStateContext.test.jsx
enZane: Probably one for length and one to check defaultLocal for both stores, because the length and the defaultLocal looks like 2 different behaviors

enZane (approved):

enZane (approved):

enZane:
LGTM!

PR #183, opened 2022-07-25
DB-3228: [next-drupal-starter] Snapshot tests

CobyPear:
## What changes were made?
- Add snapshot tests for each page route
- Tests are configured to run for each backend profile, default and umami
- Test data is added so tests can run in any environment

## Additional information

`next/image` and `next/router` need to be mocked in the snapshot tests, so we are not testing image optimization.


starters/next-drupal-starter/README.md
```diff
@@ -88,7 +89,8 @@ See our [Backend Starters](https://github.com/pantheon-systems/decoupled-kit-js/
 ### Commands
 
 This section assumes the package manager in use is `npm`. If you are not using `npm`, replace `npm` with the name of your package manager.
-To run the tests for both profiles sequentially:
+
+To run all tests for both profiles sequentially:
```
backlineint, line 93: These tests are great! (including the unit tests, which I didn't have a chance to try until now)

One thing that I'd expect to see here is an explanation of how to update the snapshot tests if there is a valid change. Wasn't clear on how to do that.

For example, I ran the tests and they passed. Then I changed the value of the H1 on the homepage header. The snapshot test failed, which is good. But since in this pretend example that is an intentional change, I'll need to know how to update the snapshot the test passes going forward.

backlineint (changes requested):

backlineint (approved):

PR #184, opened 2022-07-25
DB-3465: [next-wordpress] Block Editor Quote Styles 

enZane:
## What changes were made?
Added support for new classes in WP6
Renamed files in next-wordpress starter to use .jsx
Added classes for quotes from WP

Wordpress-kit
next-wordpress-starter

With old and new WP backends

## Additional information


starters/next-wordpress-starter/pages/pages/index.jsx
```diff
@@ -15,7 +16,7 @@ export default function PageListTemplate({ menuItems, pages }) {
         description="Generated by create next app."
       />
       <PageHeader title="Pages" />
-      <section>
+      <section className="has-luminous-vivid-amber-background-color">
```
CobyPear, line 19: Left over from testing?

CobyPear (changes requested):
Looks good!

I noticed that a long string in a pargraph can overflow the block. I think we may want to wrap here, what do you think? Do we want to keep things from overflowing their container?
 
image

starters/next-wordpress-starter/pages/pages/index.jsx
enZane: Oh yes, I forgot to revert this, thanks for the catch

enZane:
I added to the wrapper where we insert the Html from WordPress the class break-words that wrap it, is okay this approach?

CobyPear:
> I added to the wrapper where we insert the Html from WordPress the class break-words that wrap it, is okay this approach?

Works for me!

CobyPear (approved):

PR #185, opened 2022-07-25
Downgrade gatsby@4.19.0 to 4.16.0

CobyPear:
## What changes were made?
Downgrade back to gatsby 4.16.0 as there is some unknown issue when using `placeholder: TRACED_SVG`.

starter builds locally

## Additional information

### DIFF
--- a/.changeset/lazy-poets-count.md
+++ b/.changeset/lazy-poets-count.md
@@ -1,0 +1,5 @@
+---
+"@pantheon-systems/gatsby-wordpress-starter": minor
+---
+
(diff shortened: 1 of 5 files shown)


backlineint (changes requested):
@CobyPear running in dev mode I see the following warning:

warn ./.cache/root.js
Attempted import error: 'BaseContext' is not exported from '@gatsbyjs/reach-router' (imported as
'BaseContext').

Which ends up throwing a runtime error in the browser. 

I was seeing similar things when trying to pinpoint which commit introduced this regression.

backlineint (approved):
This looks to be working for me locally now.

PR #186, opened 2022-07-26
DB-3466: [next-wordpress] Block Editor Pullquote Styles

enZane:
## What changes were made?
- Added pullquote support
- Added a file for types in the plugin
- Split in separated files the constants and config from the plugin

- Wordpress-Kit

- Locally and built

## Additional information


CobyPear (changes requested):
I like the organizational changes to the config.
One style I would like to see retained from WP is the italics in the pullquote. LGTM other than that!

CobyPear (changes requested):
The paragraph is italic now which is good, but the cite seems to stay italic as well (see screenshot 1).

I also noticed the fontStyle is normal on these `quoteUtilities` except for one. Is this intended?
```
const quoteUtilities = {
...
    '.is-style-large': {
      cite: {
        fontStyle: 'italic',
        fontSize: '1.2rem',
      },
```

Also, I missed the "solid color" style for the pullquote which seems to be missing from the plugin (see screenshot 2).

image

image

CobyPear (commented):
OK Reviewing again with the correct backend.. looks great!

One thing I'm not seeing reflecting in our FE is this typography size; changing it doesn't seem to have much effect.

image

Typography 1
image


Typography 4
image

Can we up the font size instead of the background size here? Or does it make more sense to keep it how it is? @backlineint do you have an opinion here?

enZane:
We should follow the WP way that only increases the font size in the cite? Or both of them?

backlineint:
@enZane @CobyPear 

> We should follow the WP way that only increases the font size in the cite? Or both of them?

WP's styles here are kind of odd. But in general, I think we should try to follow the spirit of what WP is doing. So I'd say keep what we have now with the padding, but adjust to it also impacts the size of the cite as WP does.

CobyPear (approved):
LGTM 🚀

PR #188, opened 2022-07-28
DB-3134: Add catch all route to next-drupal-starter

CobyPear:
## What changes were made?
Added a catch all route for the next-drupal-starter so top level articles, pages, and recipes will render appropriatly.

next-drupal-starter

tested locally. Snapshot tests added for each profile/content type that is valid for the profile

## Additional information


backlineint (changes requested):
Catch all route work itself looks great, but I'm seeing a regression for decoupled preview - all routes are throwing a 500.

backlineint (approved):

PR #191, opened 2022-07-28
DB-3453: [next-wordpress] Invalid href passed to next/router

enZane:
## What changes were made?
- Added a formatter to the lib/Menus.js that take the URI and returns the pathname, and is applied when the URI starts with HTTP to fix the console error

- next-wordpress-starter

- locally and built
- with the older QA WP site and the new with WP6

## Additional information

### DIFF
--- a/.changeset/neat-frogs-dress.md
+++ b/.changeset/neat-frogs-dress.md
@@ -1,0 +1,5 @@
+---
+"@pantheon-systems/next-wordpress-starter": patch
+---
+
+Add a URI formatter in lib/Menus to fix a problem with WP6 implementation
--- a/.changeset/neat-frogs-dress.md
+++ b/.changeset/neat-frogs-dress.md
@@ -2,4 +2,4 @@
 "@pantheon-systems/next-wordpress-starter": patch
 ---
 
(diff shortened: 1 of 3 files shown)


CobyPear (changes requested):
While this is a good solution, I think there's a more straightforward path

in Menus.js
```
`
    query FooterMenuQuery {
      menu(idType: NAME, id: "Example Menu") {
        menuItems {
          edges {
            node {
              id
              path <- We just need to switch this from `uri` to `path`
              label
            }
          }
        }
      }
    }
  `
```

And then in the footer.jsx component, replace `uri` with `path`

enZane:
But than won't break the previous integration??

CobyPear:
> But than won't break the previous integration??

Not sure I understand. I tested it locally and it seemed fine.

enZane:
I mean, with the older WordPress site

CobyPear:
> I mean, with the older WordPress site

I just tested the WP5 backend and it seems there was a regression somewhere we didn't catch. I think we need to figure out if we are supporting both wp5 and wp6 with this starter, and if so, how. The `path` field is available in both 5 and 6 so it shouldn't cause a regression.

CobyPear (approved):
👍

PR #193, opened 2022-07-29
DB-3466: Add table styles

enZane:
## What changes were made?
- Added styles to render tables

- Wordpress-kit

- Locally

## Additional information

- I'm not sure about the block alignment, because in the WP block editor it looks broken 
image
but is in the PR, if we decide to don't honor it, then I can remove that behavior

### DIFF
--- a/packages/wordpress-kit/src/lib/tailwindcssPlugin.ts
+++ b/packages/wordpress-kit/src/lib/tailwindcssPlugin.ts
@@ -4,7 +4,7 @@ import { Color } from '../types/tailwindcssPlugin';
 import { mergeToConfig } from './tailwindCssPlugin/config';
 import { colorList, fontSizeList } from './tailwindCssPlugin/constants';
 
-export default plugin(function ({ addUtilities, theme }) {
+export default plugin(function ({ addUtilities, theme, addComponents }) {
   const getColor = (color: Color) =>
     theme(
       `colors.${color.themeName}`,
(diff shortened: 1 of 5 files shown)


CobyPear (changes requested):
Small tables look good. I have a large table with "Full Width" selected in WP, but the article body is forcing the table to allign-left and overflow which looks a bit strange. Is it possible to have this table be centered on the page?

Also, can the <figcaption> be center aligned?

image

starters/next-wordpress-starter/components/post.jsx
```diff
@@ -5,19 +5,16 @@ export default function Post({
   post: { title, date, featuredImage, content },
 }) {
   return (
-    <article className="prose lg:prose-xl mt-10 mx-auto">
+    <article className="prose lg:prose-xl mt-10 mx-auto max-w-screen-lg">
```
CobyPear, line 8: Let's add some padding here, probably `px-4` or `p-4`

CobyPear (changes requested):
One small request with the padding on the posts, and if you can solve the overflow gap shown in the image, but I'm not sure that was introduced in this PR. Might need to scroll the overflow on the largest table size or something.

image

packages/wordpress-kit/src/lib/tailwindcssPlugin.ts
```diff
@@ -174,6 +174,74 @@ export default plugin(function ({ addUtilities, theme }) {
...
+          maxWidth: '850px',
+        },
+      },
+      '&.alignfull': {
+        [`@media (min-width:${theme(
+          'screen.xl',
+          '1280px'
+        )})` as '@media(min-width:1280px)`']: {
+          marginLeft: 'calc(-1 * max(1rem, 10vw))',
+          marginRight: 'calc(-1 * max(1rem, 10vw))',
```
CobyPear, line 232: This is some nice wizardry. Can you add a comment or two explaining what the `calc` is doing, and also maybe the what is going on with the type coercion so we can more easily maintain this in the future?

starters/next-wordpress-starter/components/post.jsx
```diff
@@ -5,19 +5,16 @@ export default function Post({
...
       <Link passHref href="/">
         <a className="font-normal">Home &rarr;</a>
       </Link>
-      <div className="mt-12 max-w-lg mx-auto lg:grid-cols-3 lg:max-w-screen-lg">
-        {featuredImage && (
-          <div
-            className="relative w-full rounded-lg shadow-lg overflow-hidden mb-10"
-            style={{ height: "50vh" }}
-          >
+      {featuredImage && (
```
CobyPear, line 15: I am not able to see images after I pulled the latest changes.
Can we revert this section to how it was? If you want to put that outer div inside of the conditional too, that's OK.

```jsx
      <div className="mt-12 max-w-lg mx-auto lg:max-w-screen-lg">
        {featuredImage && (
          <div
            className="relative w-full rounded-lg shadow-lg overflow-hidden mb-10"
            style={{ height: "50vh" }}
          >
            <Image
              priority
              src={featuredImage.node.sourceUrl}
              layout="fill"
              objectFit="cover"
              alt={featuredImage.node.alt || "Featured Image"}
            />
          </div>
        )}
      </div>
```

CobyPear (changes requested):
Table styles are looking great! I noticed a regression with the `featuredImage`, let's revert that part of the post component if it makes sense to you. Also just a note to add some comments to the less obvious tailwind plugin code.

CobyPear (approved):
@enZane I made the changes I mentioned in the last comment.

PR #198, opened 2022-08-03
DB-3467: Add table styles

enZane:
## What changes were made?
- Change default theme colors to follow WP6 styles
- Fix on border styles for tables

- Wordpress-kit
- Locally

## Additional information

### DIFF
--- a/.changeset/curvy-birds-wink.md
+++ b/.changeset/curvy-birds-wink.md
@@ -1,0 +1,5 @@
+---
+"@pantheon-systems/wordpress-kit": patch
+---
+
+Change the default colors to follow WP6 and fix border styles in tables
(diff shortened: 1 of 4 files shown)


CobyPear (changes requested):
I think this is good, just one ask that the [`WordpressMapConfig`](https://github.com/pantheon-systems/decoupled-kit-js/blob/960e2e83e3a9655760a4260e7d18b0159fb9d5a2/packages/wordpress-kit/src/types/tailwindcssPlugin.d.ts#L59) be updated, mostly the colors. (Also we should update the name to `WordPressMapConfig`, I think they are strict about the capital P lol)
If I'm extending the config it's nice to see the names of those other colors like you have for what is there now.

CobyPear (approved):
LGTM!

PR #199, opened 2022-08-03
DB-3552: Added alias for common commands, alphabetization.

mitchellmarkoff:
## What changes were made?
Added shorter commands to build, develop, and start the starters.

decoupled-kit-js

Manually executing alias commands on local machine. 

## Additional information


package.json
```diff
@@ -13,14 +13,24 @@
   "author": "@pantheon-systems",
   "main": "index.js",
   "scripts": {
-    "test": "pnpm recursive test",
+    "build:all": "pnpm recursive run build",
```
CobyPear, line 16: This might be better constructed as `pnpm build:pkgs && pnpm build:starters` since the starters have a dependency on the packages. This would also leave out docs from being built which I think is fine for right now.

Also, I know it was there before, but we can get rid of 'run' in all of the build commands.

package.json
```diff
@@ -13,14 +13,24 @@
...
+    "ci:publish": "changeset publish",
+    "ci:version": "changeset version",
+    "dev:gatsby-wp": "pnpm --filter './starters/gatsby-wordpress-starter' develop",
+    "dev:next-drupal": "pnpm --filter './starters/next-drupal-starter' dev",
+    "dev:next-wp": "pnpm --filter './starters/next-wordpress-starter' dev",
     "generate-docs": "pnpm --filter './web/**' clear && pnpm --filter './web/**' build",
     "postinstall": "pnpm build:pkgs",
-    "ci:version": "changeset version",
-    "ci:publish": "changeset publish"
+    "serve:gatsby-wp": "pnpm -F './starters/gatsby-wordpress-starter' serve",
```
CobyPear, line 29: Let's stick with `--filter` for all of these.

CobyPear (changes requested):
Just a few little changes but overall, good stuff!

CobyPear (approved):
🥳 Grats on your first PR to the project!

PR #200, opened 2022-08-04
DB-3650: Add image block support

enZane:
## What changes were made?
- Added styles to render image blocks

- WordPress-kit

- Locally

## Additional information


CobyPear (changes requested):
Just one issue I'm seeing at the moment::
Looks like a right or left align images might have their container overlayed by another shown here

3 images on a page
Actual ⬆️

4 images on a page
Expected ⬆️

CobyPear (changes requested):
LGTM. One tiny change, can you rename the `lib/tailwindCssPlugin` dir to `lib/tailwindcssPlugin` to keep the naming consistent (lowercase css)?

I like the idea of moving the components to their own files. Maybe we should do the same for Utilities, but it can be in another ticket or PR.

enZane:
I also tried to move the utilities but I need to find a way to pass the theme function without TS problems, I'll give it another try in the gradient ticket

CobyPear (approved):
👍

PR #201, opened 2022-08-04
Changes for DB-3629 - Update Image.src to use the IMAGE_DOMAIN

lotusbaba:
## What changes were made?
Added changes to enable Image domain for next js Wordpress
	modified:   components/page.jsx
	modified:   components/post.jsx
	modified:   lib/Posts.js
	modified:   package-lock.json <-- Modified by npm install. I didn't see it being ignored in .gitignore so committed this file too in my changes

next-wordpress-starter

Yes on local env with the backend - dev-decoupled-wp6-test.pantheonsite.io

## Additional information

### DIFF
--- a/starters/next-wordpress-starter/components/page.jsx
+++ b/starters/next-wordpress-starter/components/page.jsx
@@ -6,7 +6,7 @@ import { IMAGE_URL } from "../lib/constants";
 export default function Page({
   page: { title, date, featuredImage, content },
 }) {
-  let srcUrl = getUrlPath(featuredImage.node.sourceUrl);
+  const srcUrl = featuredImage?.node?.sourceUrl && getUrlPath(featuredImage.node.sourceUrl);
 
   return (
     <article className="prose lg:prose-xl mt-10 mx-auto">
--- a/starters/next-wordpress-starter/components/page.jsx
+++ b/starters/next-wordpress-starter/components/page.jsx
@@ -4,6 +4,23 @@ import Link from "next/link";
 export default function Page({
   page: { title, date, featuredImage, content },
 }) {
+  let rxSrc, imgSrc, rxPath, imgPath;
+  let srcUrl = "";
+  try {
+    rxSrc = new RegExp(".+/(.+.+$)"); //We will get the file name
+    imgSrc = rxSrc.exec(featuredImage.node.sourceUrl)[1];
+    rxPath = new RegExp("http[s]?:\\/\\/?[^\\/]+\\/(.*[/]{1}.*/).+.+$");
+    imgPath = rxPath.exec(featuredImage.node.sourceUrl)[1];
+    if (process.env.IMAGE_DOMAIN)
+      srcUrl = "https://" + process.env.IMAGE_DOMAIN + "/" + imgPath + imgSrc;
+    else srcUrl = featuredImage.node.sourceUrl;
+  } catch (e) {
+    console.log(
+      "Regexp of image domain construction failed. This is the error: " + e
+    );
+    srcUrl = featuredImage.node.sourceUrl;
+  }
+
   return (
     <article className="prose lg:prose-xl mt-10 mx-auto">
       <h1>{title}</h1>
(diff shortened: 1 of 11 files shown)


starters/next-wordpress-starter/components/page.jsx
```diff
@@ -4,6 +4,23 @@ import Link from "next/link";
...
   page: { title, date, featuredImage, content },
 }) {
+  let rxSrc, imgSrc, rxPath, imgPath;
+  let srcUrl = "";
+  try {
+    rxSrc = new RegExp(".+/(.+.+$)"); //We will get the file name
+    imgSrc = rxSrc.exec(featuredImage.node.sourceUrl)[1];
+    rxPath = new RegExp("http[s]?:\\/\\/?[^\\/]+\\/(.*[/]{1}.*/).+.+$");
+    imgPath = rxPath.exec(featuredImage.node.sourceUrl)[1];
+    if (process.env.IMAGE_DOMAIN)
```
CobyPear, line 14: This will always be false since env vars on `process.env` are not made available to the client side unless it is prefixed with `NEXT_PUBLIC`, which is why I suggested a similar approach that the `next-drupal-starter` is taking with the `lib/constants.js` file

CobyPear (changes requested):
While this approach looks like it could be quite robust, I think there is a more straightforward and maintainable solution where we remove the `WPGRAPHQL_URL` base domain name from the `featuredImg.node.sourceUrl`.

Apologies because I think yesterday I said we need to remove the `IMAGE_DOMAIN`, but that probably caused some confusion.

See attachment for some suggestions

image

starters/next-wordpress-starter/components/page.jsx
```diff
@@ -4,6 +4,24 @@ import Link from "next/link";
 export default function Page({
   page: { title, date, featuredImage, content },
 }) {
+  let rxSrc, imgSrc, rxPath, imgPath;
```
backlineint, line 7: We should abstract this out into a reusable utility of some kind. I think there are some other things in the components folder that need this change applied, and as people build on top of this starter there will likely be even more instances where this code snippet would need to be duplicated.

backlineint (changes requested):

starters/next-wordpress-starter/components/page.jsx
lotusbaba: I'm issuing a new PR and closing this one. Added this suggestion to it with this code in a new utilities.jsx file

starters/next-wordpress-starter/lib/getUrlPath.js
```diff
@@ -0,0 +1,4 @@
+export const getUrlPath = (url) => {
+  var rxPath = new RegExp("http[s]?:\\/\\/?[^\\/]+\\/(.*[/]{1}.*/.+.+$)");
```
CobyPear, line 2:
```suggestion
  const rxPath = new RegExp("http[s]?:\\/\\/?[^\\/]+\\/(.*[/]{1}.*/.+.+$)");
```
Not much reason to use `var` with ES6

starters/next-wordpress-starter/components/post.jsx
```diff
@@ -1,9 +1,13 @@
 import Image from "next/image";
 import Link from "next/link";
+import { getUrlPath } from "../lib/getUrlPath";
+import { IMAGE_URL } from "../lib/constants";
 
 export default function Post({
   post: { title, date, featuredImage, content },
 }) {
+  let srcUrl = getUrlPath(featuredImage.node.sourceUrl);
```
CobyPear, line 9:
```suggestion
  const srcUrl = featuredImage?.node?.sourceUrl && getUrlPath(featuredImage.node.sourceUrl);
```
Here and in `page.jsx`, we can add some optional chaining and short circuit evaluation so that we only pass in `featuredImage.node.sourceUrl` if it is truthy.

CobyPear (changes requested):
Getting some build errors in cases where the featuredImage is not set. See comment for suggestion.

CobyPear (approved):
LGTM 🥳
Grats on your first (non-docs) PR!

PR #208, opened 2022-08-11
DB-3262: `lando` setup for `decoupled-wordpress-composer-managed`.

abhisekmazumdar:
## What changes were made?

Added documentation for local setup on `decoupled-wordpress-composer-managed` repo via `lando`


web/docs/Backend Starters/Decoupled WordPress/local-development.md
```diff
@@ -0,0 +1,6 @@
+# Local Development Documentation
+
+## lando setup for [decoupled-wordpress-composer-managed](https://github.com/pantheon-systems/decoupled-wordpress-composer-managed)
```
backlineint, line 3: The acceptance criteria for this ticket outlines that we have a setup for backend only, and another for a combined front end and backend. Could you add a backend only section here? That may just be the standard process for running Lando init.

web/docs/Backend Starters/Decoupled WordPress/local-development.md
```diff
@@ -0,0 +1,6 @@
...
+
+- Copy & rename [template.lando.yml](https://gist.github.com/abhisekmazumdar/649fd0937574a61becf083fc4979843a) to `.lando.yml` into the root of the project.
```
backlineint, line 5: Some adjustments the Lando template:
* For the name, proxy, and site could you use placeholders rather than a value here. People will likely want to adjust these for their site.
* Could you add a placeholder for site id as well?
* Since we don't support pnpm for the deployed sites, could you remove pnpm from this template and use npm by default instead?

backlineint: I think this also needs a little more context about what belongs in the frontend folder. In this case it wouldn't be this repository, it would be the repository for their front end project.

backlineint (changes requested):

web/docs/Backend Starters/Decoupled WordPress/local-development.md
abhisekmazumdar: I didn't understand the last comment. Do you mean these changes should be in some other place?

abhisekmazumdar:
Wrong Branch name!!

PR #210, opened 2022-08-11
DB-3651: [next-wordpress] Block Editor Gallery Styles

enZane:
## What changes were made?
- Added support for Gallery block

- Wordpress-kit

- Locally

## Additional information

### DIFF
--- a/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Gallery.ts
+++ b/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Gallery.ts
@@ -1,0 +1,149 @@
+const getFigureStyles = (mobileScreen: string) => {
+  let styles = {
+    '&.columns-default': {
+      figure: {
+        // This calc is to divide the width by the number of columns
+        // and then subtract 1rem to account for the gap between columns/rows.
+        [`@media (min-width: ${mobileScreen})`]: {
+          width: 'calc(33.33% - 1rem) !important',
+        },
+      },
+    },
+    figure: {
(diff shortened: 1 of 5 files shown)


packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Gallery.ts
```diff
@@ -0,0 +1,138 @@
...
+    },
+  };
+
+  for (let column = 1; column <= 8; column++) {
+    styles = {
+      ...styles,
+      [`&.columns-${column}`]: {
+        figure: {
+          [`@media (min-width: ${mobileScreen})`]: {
+            width: `calc(${100 / column}% - 1rem) !important`,
```
CobyPear, line 21: Can you add a comment here explaining what the `calc()` is doing?

CobyPear (changes requested):
When I make a gallery with 5 images in WP I see 3 column behavior, but the frontend representation is 2 column behavior (see image).
If I manually set the columns on the block to 3 in WP, I get the desired behavior, but I would expect it to be like that without setting it explicitly.

Actual layout showing 4 images of cats in 2 columns and a large image on the bottom, next to a smaller layout of the same images but 3 columns on top and 2 larger images on the bottom.

packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Gallery.ts
enZane: Sure it!

CobyPear (approved):
LGTM

PR #216, opened 2022-08-15
DB-3606: Add header and footer components to nextjs-kit

mitchellmarkoff:
## What changes were made?
Implementations of the Header and Footer components were added to the nextjs-kit component library. Tailwind config added to project.

nextjs-kit
decoupled-kit-js

Tested in local implementations into the starter kits. Snapshot test written.

## Additional information


packages/nextjs-kit/src/lib/interfaces.ts
```diff
@@ -0,0 +1,10 @@
+export interface Headers {
+  title: string;
+  href: string;
+}
+
+export interface Menu {
+  label: string;
+  path: string;
+  id: string;
+}
```
CobyPear, line 10: Let's move types to `src/types/index.ts`

packages/nextjs-kit/tailwind.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  content: [
+    './pages/**/*.{js,ts,jsx,tsx}',
```
CobyPear, line 3: I think this part can be removed

packages/nextjs-kit/src/components/header.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface HeaderProps {
+  headers: Headers[];
+}
+
+/**
+ * This is a Header component.
+ *
+ * @param {Props} headers holds the title and href of each route to be displayed
+ * @returns React.JSX.Element
```
CobyPear, line 13:
```suggestion
 * @returns {React.JSX.Element}
```
Wrapping the type in `{}` should help our TSDoc definitions

CobyPear (changes requested):
As I mentioned to you earlier, I think I messed up on our type exports because when I import this package to one of the starters, I don't see any type info on our component, and the "Go to Definition" feature of Vscode isn't working either.

I think I will open another ticket for that stuff. I left some comments for these components here that should be addressed tho. In addition to those comments, you can remove the `Test` and `Test2` components from the package now that we have some real implementations.

Also some pickyness:
- Add the ticket number to the PR. You had one but it was a different ticket number.
- Format commits with the ticket number as well like `DB-3606: {some message}`
- Also, less important but if branches could also start with the ticket number it would be helpful 🙇

packages/nextjs-kit/src/components/footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const FooterMenu = () => (
+    <nav className="flex flex-col max-w-lg mx-auto lg:max-w-screen-lg">
+      <ul>
+        {menuItems.map(({ label, path, id }) => {
+          return (
+            <li
+              key={id}
+              className="list-disc text-blue-300 hover:text-blue-100 ml-3"
+            >
+              <Link href={`/posts${path}`}>
```
CobyPear, line 27:
```suggestion
              <Link href={`/${path}`}>
```
I thought this was only in the snapshot component before so I didn't catch this.
You may also need to check for leading `/` on the path and remove it to avoid passing in `//path`.

packages/nextjs-kit/src/components/footer.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import Link from 'next/link';
+import { Menu } from '../index';
+
+interface MenuProps {
```
CobyPear, line 5: Be more specific here
```suggestion
interface FooterMenuProps {
```

packages/nextjs-kit/src/components/footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param {Props} menuItems holds the title, href, and id of each route to be displayed
+ * @param {Props} children  copyright display and information to be passed in by the user
+ * @returns {React.JSX.Element}
+ */
+const Footer: React.FC<MenuProps> = ({ menuItems, children }: MenuProps) => {
+  const FooterMenu = () => (
+    <nav className="flex flex-col max-w-lg mx-auto lg:max-w-screen-lg">
+      <ul>
+        {menuItems.map(({ label, path, id }) => {
```
CobyPear, line 21: When I tried to use this component by passing in our "footerMenu" coming back from Drupal, I don't see any of the content: 
image

Also if possible, retain the logic that makes a menu item that has a parent indented like so:

- parent item 1
  - child item 1
  - child item 2
- parent item 2

packages/nextjs-kit/src/components/header.tsx
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+const Header: React.FC<HeaderProps> = ({ headers }: HeaderProps) => {
+  return (
+    <div className="my-0 pt-10 px-5 text-xl">
+      <nav>
+        <ul className="flex flex-row flex-wrap sm:flex-nowrap list-none justify-between max-w-screen-sm mx-auto">
+          {headers.map(item => (
+            <li
+              className={`${item.href === '/' ? 'mr-auto' : 'mx-4'}`}
```
CobyPear, line 23: This works for  us, but I wonder if we can be more robust here. Maybe we can take in an "align" prop which defaults `right`, so that you can determine which side you want each menu item to be on.

packages/nextjs-kit/src/components/header.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import Link from 'next/link';
+import { Headers } from '../index';
+
+interface HeaderProps {
+  headers: Headers[];
```
CobyPear, line 6: Also rename the prop, probably something like `navItems`

packages/nextjs-kit/src/components/header.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * This is a Header component.
+ *
+ * @param {Props} headers holds the title and href of each route to be displayed
```
CobyPear, line 12:
```suggestion
 * @param {HeaderProps} headers holds the title and href of each route to be displayed
```

packages/nextjs-kit/src/components/header.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {React.JSX.Element}
+ */
+
+const Header: React.FC<HeaderProps> = ({ headers }: HeaderProps) => {
```
CobyPear, line 16: I am starting to think we should be explicit on the return of the function signature too:
```suggestion
const Header: React.FC<HeaderProps> = ({ headers }: HeaderProps): JSX.Element => {
```

CobyPear (changes requested):
Some things I missed yesterday.

Noticed that the `pnpm-lock.yaml` has no significant changes (that I could see) other than single quotes -> double quotes. We have a ticket coming up to verify our linting and stuff. For now, if you could remove the `pnpm-lock.yaml` from this commit, or at least if you could undo that formatting, I think it will prevent a back-and-forth with the quotes on that file, and when we fix/verify the linting I'll make sure it has formatting we can all agree on.

packages/nextjs-kit/src/components/footer.tsx
mitchellmarkoff: Sorry I should have left a comment about this. Our FooterMenu items in the WP starter are named differently than in the Drupal starter so changes need to be made to the object passed in one of the starters for it work with both. I imagined this work would take place in another ticket as it is not related to the components here but rather how the starter kits build these objects.

Naming is `label, path, id` vs `title, url, id`

CobyPear: Ah I see. In that case these should be a bit more generic. I guess I would expect equal work to implement this component in `next-drupal` and `next-wordpress`.

There's an opportunity here to combine these two interfaces here as well. For example:

```
export interface LinkProps {
  linkText: string;
  href: string;
}

export interface FooterMenuItem extends LinkProps {
  id: string;
}
```

This way, the header component's prop types can refer to the more generic `LinkProps` interface, which the footer menu is essentially an extension of, then we can have the same props `href` and `linkText` (maybe a better name for this) in both components, giving us some cohesion.

packages/nextjs-kit/src/components/header.tsx
```diff
@@ -0,0 +1,42 @@
...
+}: HeaderProps): JSX.Element => {
+  return (
+    <div className="my-0 pt-10 px-5 text-xl">
+      <nav>
+        <ul className="flex flex-row flex-wrap sm:flex-nowrap list-none justify-between max-w-screen-sm mx-auto">
+          {navItems.map(item => {
+            item.align = item.align !== undefined ? item.align : 'right';
+            return (
+              <li
+                className={`${item.align === 'right' ? 'mr-auto' : 'mx-4'}`}
```
CobyPear, line 27: This doesn't seem to be aligning the item left or right like I would expect. Let's remove this for now and maybe we can add something like this later.

packages/nextjs-kit/src/components/header.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface HeaderProps {
+  navItems: HeaderData[];
+}
+
+/**
+ * This is a Header component.
+ *
+ * @param {HeaderProps} navItems holds the title, href, and alignment of each route to be displayed
+ * @returns {React.JSX.Element}
```
CobyPear, line 13:
```suggestion
 * @returns {JSX.Element} A header with a navigation menu
```
JSX is its own class. Also add a description of the component that is being returned

packages/nextjs-kit/src/types/index.ts
```diff
@@ -0,0 +1,15 @@
+export interface HeaderData {
```
CobyPear, line 1: Still not wild about this interface name. Let's match it up with the name of the prop which in this case is `navItems`, so we can call this `NavItems`.

Also, we can compose our types a bit more so that instead of `FooterProps`, we have `LinkProps`, and then `FooterMenuItem` can extend `LinkProps` and you can use `LinkProps[]` as the type for the `HeaderProps.navItems`

packages/nextjs-kit/src/types/index.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export interface FooterProps {
+  linkText: string;
+  href: string;
+}
+type Parent = 'parent' | 'parentId';
+
+export type FooterMenuItem = FooterProps & {
+  [key in Partial<Parent>]?: string | null;
```
CobyPear, line 14: I don't think the partial actually does what we thought here. The `?` should be enough to make it optional.
```suggestion
  [key in Parent]?: string | null;
```

packages/nextjs-kit/src/components/footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+const Footer: React.FC<FooterMenuProps> = ({
+  footerMenuItems,
+  children,
+}: FooterMenuProps) => {
+  const FooterMenu = () => {
+    const menuArr = [];
+    if (footerMenuItems) {
+      // some not so great code to account for nested menu elements
+      for (let i = 0; i < footerMenuItems.length; i++) {
+        if (footerMenuItems[i + 1] && footerMenuItems[i + 1].parent) {
```
CobyPear, line 26: this wouldn't work for the WordPress data that includes `parentId`. This might be a good opportunity to try a type predicate:

```ts
const hasParent = (item: FooterMenuItem): item is FooterMenuItem =>
  item.parentId ? true : item.parent ? true : false;
```

Then, in the loop we can check if `parent` exists:

```ts
const itemHasParent = hasParent(footerMenuItems[i + 1]);
```

packages/nextjs-kit/src/components/footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+  children: JSX.Element;
+}
+
+/**
+ * This is a Footer component.
+ *
+ * @param {Props} footerMenuItems holds the linkText, href, and parent of each route to be displayed
+ * @param {Props} children  copyright display and information to be passed in by the user
+ * @returns {React.JSX.Element}
+ */
```
CobyPear, line 16: I think there's actually just the one param (called `props`) which in this case is of type `FooterMenuProps`
In this way we need to call out the first param, then the nested ones. See https://typedoc.org/tags/param/
```suggestion
/**
 * This is a Footer component.
 *
 * @param {FooterMenuProps} props - The props needed for the footer component
 * @param {FooterMenuItem[]} props.footerMenuItems - An array of footer menu items
 * @example
 * ```
 * const footerMenuItems = [
 *  {
 *   href: '/',
 *   title: 'Home',
 *  },
 * {
 *  href: '/articles',
 *  title: 'Articles',
 *  parent: 'home',
 * },
 * ...
 *  ]
 * ```
 * @param {ReactChildren} props.children
 * @returns {JSX.Element} A footer component with a nav menu
 */
```

CobyPear (changes requested):
I think our compositional types got a bit lost after our last pairing session. Also the align doesn't seem to be working, maybe it's not worth worrying about for now.

CobyPear:
Oh one other thing I keep seeing during the build and forgetting to add a comment on–add `next/link` to the `rollup.config.js` globals so we can get rid of this warning.


```shell
src/components/footer.tsx, src/components/header.tsx, src/lib/testLib.ts → dist/es...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
next/link (imported by src/components/header.tsx, src/components/footer.tsx)
created dist/es in 791ms

umd-entrypoint.ts → dist/umd/nextjs-kit.umd.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
next/link (imported by src/components/header.tsx, src/components/footer.tsx)
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
next/link (guessing 'Link')
created dist/umd/nextjs-kit.umd.js in 491ms
```

packages/nextjs-kit/src/components/footer.tsx
mitchellmarkoff: Was spending too much time on the CR revisions for the CSS on the child component so pushed without a solution. Can maybe circle back on this next week?

CobyPear (approved):
There were a few comments that needed updating and I moved the type predicate (`hasParent`) to the `types` folder, and I think I got the menu css figured out (for now at least, we should make this better in the future)

PR #221, opened 2022-08-18
DB-3263: lando setup for decoupled-wordpress-composer-managed.

abhisekmazumdar:
## What changes were made?
Added documentation for local setup on decoupled-wordpress-composer-managed repo via lando


backlineint (approved):
@abhisekmazumdar made a few revisions here which should be considered as you work on the comparable Drupal ticket.

PR #223, opened 2022-08-18
DB-3547: API data sort helper functions

mitchellmarkoff:
## What changes were made?
Added two API helper functions `sortDate` and `sortChar` to handle the sorting of objects passed in by the user.

nextjs-kit

Unit test written, kit helper functions were imported into the starters and tested locally.

## Additional information


packages/nextjs-kit/src/lib/sortDate.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param {Array<Record<string, string | number>>} obj object to sort
+ * @param {string} key to sort object by
+ * @param {string} direction to sort object by
+ * @returns {Array<Record<string, string | number>>} the sorted object that was passed in
+ * @remarks
+ * This is a function to sort any date filed of an object on a specific key in a direction of the users choice.
+ *
+ * Note:
+ * There is a named and default export so that we can
+ * support submodules for the ESModule build and a single entrypoint for the UMD build
```
CobyPear, line 12: This note can be removed, it was mostly an explainer on the test component

packages/nextjs-kit/src/lib/sortChar.ts
```diff
@@ -0,0 +1,30 @@
...
+ * This is a function to sort any character object on a specific key in a direction of the users choice.
+ *
+ * Note:
+ * There is a named and default export so that we can
+ * support submodules for the ESModule build and a single entrypoint for the UMD build
+ */
+export const sortChar = (
+  obj: Array<Record<string, string | number>>,
+  key: string,
+  direction: string
```
CobyPear, line 17: Can we make these parameters an object?
also here is a case where an `enum` may be a good idea for the `direction` type, also I would recommend shortening the names from ascending and descending to `'ASC'` and `'DESC'`.

packages/nextjs-kit/src/__tests__/sortDate.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  {
+    date: '1996-06-17T15:12:39+00:00',
+  },
+];
+
+describe('sortChar()', () => {
+  it('should return the unsorted numbers in a sorted ascending order', () => {
+    expect(sortDate(unsortedDates, 'date', 'ascending')).toEqual(sortedAsc);
+  });
+  it('should return the unsorted numbers in a sorted descending order', () => {
```
CobyPear, line 51: The labels here reference `sortChar` and 'numbers' but this looks like the `sortDate()` tests

packages/nextjs-kit/src/lib/sortDate.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const sortDate = (
+  obj: Array<Record<string, string>>,
+  key: string,
+  direction: string
+) => {
```
CobyPear, line 18: Also make these params an object. Maybe create an interface that can be shared between these helpers?

CobyPear (changes requested):
A few picky things but it looks good otherwise. I see it was missed in the acceptance criteria, but we should also import this helper into the Drupal starter and utilize it to get the data sorted in a consistent way.

packages/nextjs-kit/src/types/index.ts
```diff
@@ -0,0 +1,14 @@
...
+export enum DESC {
+  DESC = 'DESC',
+  desc = 'desc',
+}
+
+export interface DataToSort {
+  data: Array<Record<string, string | number>>;
+  key: string;
+  direction: string;
+}
```
CobyPear, line 14: I think this can be cleaned up a bit.

Note that we can use `Type[]` instead of `Array<Type>`, the latter of which is more difficult to read imo.
```suggestion
export interface DataToSort {
  data: Record<string, string | number>[];
  key: string;
  direction: 'asc' | 'ASC' | 'desc' | 'DESC';
}
```

packages/nextjs-kit/src/lib/sortChar.ts
```diff
@@ -0,0 +1,41 @@
...
+      a[sortObj.key] > b[sortObj.key]
+        ? -1
+        : a[sortObj.key] < b[sortObj.key]
+        ? 1
+        : 0
+    );
+  }
+
+  return sortObj.data;
+};
```
CobyPear, line 39: This can be a bit less verbose by changing which object is on either side of the `<` operator instead of defining two callbacks:

```suggestion
 export const sortChar = (sortObj: DataToSort) => {
  sortObj.data.sort((a, b) => {
    const direction = sortObj.direction.toLowerCase();
    const left = direction === 'asc' ? b : a;
    const right = direction === 'asc' ? a : b;
      return left[sortObj.key] > right[sortObj.key]
          ? -1
          : left[sortObj.key] < right[sortObj.key]
              ? 1
              : 0
  });
  return sortObj.data;
};
```

packages/nextjs-kit/src/lib/sortChar.ts
```diff
@@ -0,0 +1,41 @@
+import { DataToSort, ASC, DESC } from '../types';
+
+/**
+ *
+ * @param {Array<Record<string, string | number>>} obj object to sort
+ * @param {string} key to sort object by
+ * @param {string} direction to sort object by
```
CobyPear, line 7: For the params, we can use the types from our interface instead of declaring the primitives here:

```suggestion
 * @param {DataToSort['data']} sortObj.data - The data sort
 * @param {DataToSort['key']} sortObj.key - Key to sort object by
 * @param {DataToSort['direction']} sortObj.direction - Direction to sort the data
```

packages/nextjs-kit/src/lib/sortChar.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @returns {Array<Record<string, string | number>>} the sorted object that was passed in
```
CobyPear, line 8: Also here, we can omit the `Array<T>` type for `Type[]`
```suggestion
* @returns {Record<string, string | number>[]} An array of data sorted by the given key and direction
```

packages/nextjs-kit/src/lib/sortDate.ts
```diff
@@ -0,0 +1,42 @@
+import { DataToSort, ASC, DESC } from '../types';
+
+/**
+ *
+ * @param {Array<Record<string, string | number>>} obj object to sort
+ * @param {string} key to sort object by
+ * @param {string} direction to sort object by
+ * @returns {Array<Record<string, string | number>>} the sorted object that was passed in
```
CobyPear, line 8: Similar to comments on `sortChar`'s TSDoc comment

packages/nextjs-kit/src/lib/sortChar.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @remarks
+ * This is a function to sort any character object on a specific key in a direction of the users choice.
+ *
+ * Note:
+ * There is a named and default export so that we can
+ * support submodules for the ESModule build and a single entrypoint for the UMD build
```
CobyPear, line 14: This note should be removed

packages/nextjs-kit/src/lib/sortDate.ts
```diff
@@ -0,0 +1,42 @@
...
+        return -1;
+      }
+      if (dateA.getTime() < dateB.getTime()) {
+        return 1;
+      }
+      return 0;
+    });
+  }
+  return sortObj.data;
+};
```
CobyPear, line 40: This can be similarly shortened like `sortChar`. Also, if we are going to use ternaries like in `sortChar`, we should use them here too. Whichever you pick (if statements or ternaries), pick one and stick with it. This is something we can discuss in our code styles meeting if you feel strongly one way or the other :)

starters/next-drupal-starter/pages/index.jsx
```diff
@@ -94,6 +93,8 @@ export async function getServerSideProps(context) {
       throw new Error(
         "No articles returned. Make sure the objectName and params are valid!"
       );
+    } else {
```
CobyPear, line 96: Not sure about sticking this in an `else`. Also, are we passing the sorted articles into the props? Does simply calling this function on an array mutate the original array?  Perhaps we should create a new array and return that one if that's the case so we can avoid side-effects.

CobyPear (changes requested):

packages/nextjs-kit/src/lib/sortChar.ts
```diff
@@ -0,0 +1,29 @@
+import { DataToSort } from '../types';
+
+/**
+ *
+ * @param {DataToSort['data']} sortObj.data - The data sort
+ * @param {DataToSort['key']} sortObj.key - Key to sort object by
+ * @param  {DataToSort['direction']} sortObj.direction - Direction to sort the data
+ * @returns {Record<string, string | number>[]} An array of data sorted by the given key and direction
```
CobyPear, line 8: OK I was very wrong here, apologies!
We don't need the type at all when we are writing in TypeScript with TSDoc.
I was confused because we do need it with JSDoc.
My bad! I can do the change for these.

starters/next-drupal-starter/pages/index.jsx
```diff
@@ -90,6 +89,12 @@ export async function getServerSideProps(context) {
...
 
+    const sortedArticles = sortChar({
+      data: articles,
+      key: "title",
+      direction: "asc",
+    });
+
     if (!articles) {
       throw new Error(
         "No articles returned. Make sure the objectName and params are valid!"
```
CobyPear, line 100: I would expect the `if (!articles)` check to happen before passing articles in to `sortChar` since `sortChar` will likely throw an error if articles is undefined.

CobyPear (changes requested):
Check the snapshots you uploaded, they don't have any articles in them!

CobyPear (approved):

PR #230, opened 2022-08-25
DB-1775:Replace gatsby-wordpress-starter css style sheet with tailwind

enZane:
## What changes were made?
- Removed custom CSS
- Removed WP CSS
- Added Tailwind-WordPress plugin
- Replaced old CSS
- Moved tailwind to devDependencies
- Reset tailwind config
- Fixed some missing classes

- Gatsby-WP

- Locally

## Additional information


CobyPear (changes requested):
One small bug with the page links on the `/posts` page, I think instead of a `?` on line 119 of `gatsby-node.js`, there needs to be a `/` instead, which is what the homepage is doing and it seems to be working.

The only style change I would request is maybe some padding between the link and the edge of the screen here for smaller screens? Also for the Previous Page link
Screenshot of the Posts page with an arrow pointing to the edge of the screen near the Next page link

PR #231, opened 2022-08-26
DB-3604: Add Pagination Component

mitchellmarkoff:
## What changes were made?

- Abstracted and refactored the pagination component that is currently in the `next-drupal-starter` to be used in the `nextjs-kit`
- Fixed bug where the query param in the browser doesn’t change on the FIRST click of a button after navigating to a new shallow route

nextjs-kit

Snapshot test written. Component implemented into the next-drupal-starter locally. 

## Additional information


packages/nextjs-kit/src/types/index.ts
```diff
@@ -21,3 +21,13 @@ export type FooterMenuItem = LinkProps & {
...
+
+export const isNumber = (item: number | null | undefined): item is number =>
+  typeof item === 'number';
+export interface PaginatorObj {
+  id: string;
+  title: string;
+  body: {
+    value: string;
+  };
+}
```
CobyPear, line 33: This type is fine for our default RederData component, but if the user passes in their own component with data that does not fit this shape, I'm worried this will cause problems.

Can we make the type for the data more dynamic here? Possibly use a generic so that a user can pass in a type of data that we don't necessarily know about.

packages/nextjs-kit/src/components/paginator.tsx
```diff
@@ -0,0 +1,257 @@
...
+        {Component ? <Component currentItems={currentItems} /> : <RenderData />}
+      </section>
+      <div className="sticky lg:bottom-12 bottom-4">
+        <RenderButtons />
+      </div>
+    </div>
+  );
+};
+
+export default React.memo(Paginator);
```
CobyPear, line 257: I'm not sure if it makes sense to memoize the whole component, but maybe pieces of the component like the buttons or the RenderedData itself

packages/nextjs-kit/src/components/paginator.tsx
```diff
@@ -0,0 +1,257 @@
...
+ * @param props.data - An array of paginator objects
+ * @param props.itemsPerPage - How many items to display per page
+ * @param props.breakpoints - Breakpoints has 3 properties: start, end, and add. Set to {} for no breakpoint.
+ * start: where to start the breakpoint
+ * end: where to end the breakpoint
+ * add: how many buttons to add when the breakpoint is clicked
+ * ***
+ * note: (`add` * x) + `start` = `end` where x is a number of clicks it takes to fill in all of the buttons
+ * For example: If there are 25 buttons and the start = 5 and end = 25, then add should be 5 or 10.
+ * ***
```
CobyPear, line 17: This can probably be refactored into an `@example` and/or `@remarks` tag

packages/nextjs-kit/src/components/paginator.tsx
```diff
@@ -0,0 +1,257 @@
...
+                </p>
+              )}
+            </article>
+          );
+        })}
+      </React.Fragment>
+    );
+  };
+
+  const handlePageClick: React.MouseEventHandler<HTMLButtonElement> = event => {
```
CobyPear, line 132: here is another potential good case for a type predicate to check if the event.target is an instance of an Element, instead of relying on `as`, which can be troublesome.

CobyPear (changes requested):
Good stuff. There are a few bugs to clean up, and I think we can be more dynamic with the type of the data here since it's possible we won't know the shape until the user is implementing the component.

## 1
Seeing a rollup warning while building the package that can be resolved:

```
(!) Unresolved dependencies
│ packages/nextjs-kit build: https://rollupjs.org/guide/en/#warning-treating-module-as-external-depend
│ packages/nextjs-kit build: next/router (imported by src/components/paginator.tsx)
│ packages/nextjs-kit build: (!) Missing global variable name
│ packages/nextjs-kit build: Use output.globals to specify browser global variable names corresponding
│ packages/nextjs-kit build: next/router (guessing 'router')
```

## 2
When I use the component on a page, I don't see any styles for the buttons or anything?

image

## 3
I am also seeing a bug when implementing the component with routing and SSR: if for example I am on page 9 `localhost:3000/test/9` and I refresh, I expect to stay on the same page, but instead the route changes to `localhost:3000/test/1`, and the content also goes back to page 1.

packages/nextjs-kit/src/components/paginator.tsx
mitchellmarkoff: I used `as` here not to check if event.target is an instance of element but rather because the EventTarget type that we pass in does not automatically inherit from the Element type. [Here](https://www.designcise.com/web/tutorial/how-to-fix-property-does-not-exist-on-type-eventtarget-typescript-error) is a good blog post I found explaining the issue but I think it may be fine to use `as` in this case? Let me know your thoughts.

packages/nextjs-kit/src/components/paginator.tsx
mitchellmarkoff: Removed memoization of the whole component. I played around with profiler to test memoization on different pieces like the buttons and saw no performance increase so leaving memoization out of this component.

mitchellmarkoff:
After our conversation and some research on generics in Typescript I decided to go with the option of removing `<RenderData />` from the component. Including it would have led to some messy declarations and overall I'm not sure how helpful providing a generic data rendering component would be since incoming data will vary widely.

Add `"./node_modules/@pantheon-systems/nextjs-kit/dist/**/*.js"` to the content array in the tailwind config of each starter to observe this components styles.

packages/nextjs-kit/src/components/paginator.tsx
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * @param props.routing If true, shallow routing will be enabled. Check the examples/pagination route to see it in action
+ * @param props.Component React Component that takes in currentItems as props and maps over them.
+ * currentItems is a subset of data, so any component that works for data will work here.
+ * @see {@link https://github.com/pantheon-systems/decoupled-kit-js/tree/canary/starters/next-drupal-starter/pages/examples/pagination/[[...page]].js} for an example implementation
+ * @returns Component with data rendered by the passed in Component and page buttons
+ */
+
+interface PaginationProps {
+  data: Array<object>;
```
CobyPear, line 29: What I meant by using a generic was more for this prop, not the component. `Array<object>` doesn't give much type safety (and like we discussed in the past, it can be represented as `object[]`). Instead, we can take a generic from the user and have that be the type of the data.

```typescript
interface PaginationProps<Type> {
  data: Type[];
...
}

// then the component:
const Paginator = <Type extends object>({
  data,
  breakpoints,
  itemsPerPage,
  routing,
  Component,
}: PaginationProps) => {
...
// and as the type for current items
  const [currentItems, setCurrentItems] = useState<Type[]>([]);
```

packages/nextjs-kit/src/components/paginator.tsx
```diff
@@ -0,0 +1,234 @@
...
+ * note: (`add` * x) + `start` = `end` where x is a number of clicks it takes to fill in all of the buttons
+ * For example: If there are 25 buttons and the start = 5 and end = 25, then add should be 5 or 10.
+ * ***
+ *
+ * @param props.routing If true, shallow routing will be enabled. Check the examples/pagination route to see it in action
+ * @param props.Component React Component that takes in currentItems as props and maps over them.
+ * currentItems is a subset of data, so any component that works for data will work here.
+ * @see {@link https://github.com/pantheon-systems/decoupled-kit-js/tree/canary/starters/next-drupal-starter/pages/examples/pagination/[[...page]].js} for an example implementation
+ * @returns Component with data rendered by the passed in Component and page buttons
+ */
```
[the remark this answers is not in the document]
CobyPear: Also, make sure the TSDoc comment lives directly above the thing you are describing.

packages/nextjs-kit/src/components/paginator.tsx
CobyPear: True, but we could use a type predicate to check if something is an HTML element and avoid the type coercin, since we will know forsure what type we have.
example:

```typescript
// type predicate that determines an element is an HTMLElement:
  const isHTMLElement = (element: unknown): element is HTMLElement => {
    return element instanceof HTMLElement;
  };
```

Then in the `handlePageClick`:

```typescript
// return early if we don't have the right element type
if (!isHTMLElement(event.target)) return;
  // this should not throw typescript errors because it knows it is an HTMLElement at this point.
    if (event.target.id === 'next-btn') {
...
```

packages/nextjs-kit/src/types/index.ts
```diff
@@ -21,3 +21,13 @@ export type FooterMenuItem = LinkProps & {
...
+
+export const isNumber = (item: number | null | undefined): item is number =>
+  typeof item === 'number';
+export interface PaginatorRenderObj {
+  id: string;
+  title: string;
+  body: {
+    value: string;
+  };
+}
```
CobyPear, line 33: I think this can be removed?

CobyPear (changes requested):
Still some TS stuff to sort out; almost there.

packages/nextjs-kit/src/__tests__/snapshotTests/paginator.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+  });
+  it('back button is disabled on first page', () => {
+    render(
+      <TestComponentToRender
+        breakpoints={{ start: 6, end: 12, add: 6 }}
+        itemsPerPage={10}
+      />
+    );
+    expect(
+      (document.getElementById('back-btn') as HTMLInputElement).disabled
```
mitchellmarkoff, line 101: Let me know if there is a better way to do this. Getting to the disabled key seemed to bring some type errors. Using a type predicate here like we have in other places didn't seem eliminate the need for `as`.

CobyPear: I think it's OK for the tests, although here I would expect the element to be an [`HTMLButtonElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement)

CobyPear (changes requested):
See the one comment. Also the changes from our conversation regarding using the Paginator component directly in the tests and using the PaginationProps interface from the component so we can test the tyeps as well.

packages/nextjs-kit/src/__tests__/snapshotTests/paginator.test.tsx
```diff
@@ -83,66 +57,126 @@ const RenderCurrentItems: React.FC<PaginationItemProps> = ({
...
+            <Paginator
+              data={examplePaginationData}
+              itemsPerPage={10}
+              breakpoints={{ start: 6, end: 12, add: 6 }}
+              routing={true}
+              Component={RenderCurrentItems}
+            />
+          </section>
+        </main>
+      </div>
```
CobyPear, line 73: I don't think you need the surrounding divs. It's fine to just render the `<Paginator />` here.

packages/nextjs-kit/src/__tests__/snapshotTests/paginator.test.tsx
```diff
@@ -83,66 +57,126 @@ const RenderCurrentItems: React.FC<PaginationItemProps> = ({
...
     );
     fireEvent.click(screen.getByText('>'));
     expect(asFragment()).toMatchSnapshot();
```
CobyPear, line 162: I would expect this to fail. If the next button was hit we would need to re-render the snapshot and see what is on page 2.

It might also be worth [adding a `spyOn`](https://vitest.dev/api/#vi-spyon)to the router mock to see if `push` was called, that way if routing breaks in the component the tests will let us know.

CobyPear (changes requested):
I think the extra markup around the paginator can be removed for the tests.
Also I would expect at least one test to use `routing={false}`. You can spy on the mocked router with that test as well and it should not be called in that case.

Another test that could be nice is making sure clicking a number renders that pages's content.

Also I'm not 100% on this, but if you  click on 'next' with the `fireEvent.click()` I would expect you to need to take another snapshot, before the `expect`, otherwise we are simulating changing pages but not actually changing the content in the snapshot.

Happy to discuss that last part more if it doesn't make sense, or pair on this.

CobyPear (approved):
🥳 :shipit:

PR #240, opened 2022-09-01
DB-3021: Adjustment to local development lando template.

abhisekmazumdar:
## What changes were made?
Some changes to the local development lando template.

### DIFF
--- a/web/docs/Backend Starters/Decoupled Drupal/local-development.md
+++ b/web/docs/Backend Starters/Decoupled Drupal/local-development.md
@@ -31,6 +31,9 @@ excludes:
   - frontend/node_modules
   - node_modules
 
+env_file:
+  - frontend/.env
+
 services:
   frontend:
     type: node:16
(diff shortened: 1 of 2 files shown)


backlineint (commented):
@abhisekmazumdar I'm a little confused by this change. Is there a reason that you took this approach rather than what is outlined in the readme and the developer docs, which suggests using .env.development.local?

https://github.com/pantheon-systems/decoupled-kit-js/blob/canary/web/docs/Frontend%20Starters/Next.js/Next.js%20%2B%20Drupal/setting-environment-variables.md
https://github.com/pantheon-systems/decoupled-kit-js/tree/canary/starters/next-drupal-starter

I'd imagine this approach to sourcing envars via lando works, but it has the disadvantage of requiring a `lando rebuild` if they are changed. `.env.development.local` will be sourced every time the node process starts, which should make testing changes easier.

abhisekmazumdar:
My bad I was not aware that `.env.development.local` will load. Closing this PR as this is not required.

PR #241, opened 2022-09-01
DB-3793: [nextjs-kit] Fix `nextjs-kit` bundle

CobyPear:
## What changes were made?
- Remove rollup config, moved to the vite config `build.rollupOptions`
- Removed submodules/code splitting (I don't think I'm using the right term here, but the thing where you can import from `packageName/subpath`. Now all imports are named.
- Added tailwindcss stylesheet to the bundle, which should be imported when consuming this library
- 

`nextjs-kit`, also updated imports in `next-drupal-starter`

Tested locally by successfully using a component in a starter including its css

## Additional information


packages/nextjs-kit/src/components/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as Header } from './header';
```
backlineint, line 1: Super minor, but why not make these named imports in header.tsx and footer.tsx themeselves?

backlineint (approved):

packages/nextjs-kit/src/components/index.ts
CobyPear: Open to changing it. I was following a multitude of guides at the same time and might have left some strange portions in there.

PR #249, opened 2022-09-01
Exit pre mode when doing a main release

CobyPear:
## What changes were made?
Updates to the main release workflow. Since we are doing prereleases from the default branch (whoops), we need to exit prerelease mode before releasing on the `latest` dist tag. This _should_ achieve that by exiting pre mode before we run `pnpm changeset version`. We then enter pre mode again before all is said and done so we don't have to do so manually.

Tested exiting and entering pre mode while on the main branch and running `version` in between

## Additional information


.github/workflows/release-and-split.yml
```diff
@@ -33,11 +33,11 @@ jobs:
       - name: install dependencies
         run: pnpm install
       - name: make script runnable
-        run: chmod +x ".github/scripts/update-lockfiles.sh"
```
backlineint, line 36: @CobyPear could you elaborate on why we don't have to force the lock files to be updated anymore?

CobyPear: It wasn't really working anyway, so I decided to remove it. I think this is something we should take on manually when we have dependency changes to a `package.lock`, the `package-lock.json` should have updates as well, at least until we can get rid of the lockfile.

I think there is a way to get it working in CI, but it needs to either 

1. happen after the publish so npm can find the new version and update the lockfile without errors
2. or, we can write a script to flip the version #s this without calling `npm i`, so then the new lockfiles will be included in the Release PR.

backlineint (approved):

PR #258, opened 2022-09-06
DB-3279: Update WordPress create new project steps.

abhisekmazumdar:
## What changes were made?
Update the steps needed to create a Decoupled WordPress project using the Composer managed repo and its respective upstream.


backlineint (changes requested):
@abhisekmazumdar we don't support build tools at the moment, so could you remove the references to build tools and instead change this just to include the contents of the dashboard upstream/terminus section?

backlineint (approved):

PR #260, opened 2022-09-06
DB-3845: Standardize configs

CobyPear:
## What changes were made?
- Added a new workspace, `configs` which houses any config to be shared across packages
- Added `eslint-config` to the `configs` workspace, because ESLint has specific naming conventions when exporting a config as an npm package
- Update the packages to use these configs
- Update the prettier config with our discussed defaults
- Split the eslint config into a default (base), and `react` which extends the default config and adds the react specific plugins.
- Add a lint-staged config. There is no official way to extend a config so we are just importing the JS and re-exporting. (lint-staged has an open issue for this, this could be a nice way to contrib back to that project)
- Add a .prettierignore to the root, and use that when running `prettier` in the packages.

The starters have not been touched on this PR. Do we want to add `prettierrc` files to the next starters? If so, we should update the gatsby-wp-starter `.prettierrc` to match.

Also, changing spaces -> tabs will result in a lot of files changed. I will open another PR for this
 
 
`packages/**`, `configs`, `configs/eslint-config`

Tested `prettier`, `eslint`, and `lint-staged` commands separately for each package.

## Additional information


backlineint (approved):
In general this looks great, and the approach to dividing up the eslint configs makes a lot of sense considering Vue is on the horizon.

I had some issues seeing changes to prettier config being applied when formatting - turns out re-opening my VS code project was necessary for them to be applied.

backlineint:
> The starters have not been touched on this PR. Do we want to add prettierrc files to the next starters? If so, we should update the gatsby-wp-starter .prettierrc to match.

@CobyPear I do think that makes sense, and that approach will make it easier for people to change the config if necessary. For the Gatsby config maybe we should merge ours and theirs. We should probably keep the few Gatsby specific conventions.

PR #262, opened 2022-09-06
DB-3493: Preview Mode Ribbon

enZane:
## What changes were made?
- Added ribbon displayed on preview mode
- Now the `api/clear-preview` goes back to the same page

- Next-drupal-starter

- Locally

## Additional information


starters/next-drupal-starter/pages/pages/[...alias].jsx
```diff
@@ -106,11 +106,14 @@ export async function getServerSideProps(context) {
     });
   });
 
+  console.log("preview", context.preview);
```
Comment (CobyPear) [inline starters/next-drupal-starter/pages/pages/[...alias].jsx:109]: stray console log :)

starters/next-drupal-starter/components/preview-ribbon.jsx
```diff
@@ -0,0 +1,32 @@
+import { useState } from "react";
+
+export default function PreviewRibbion() {
```
CobyPear, line 3: This is great! Eventually we should move this to the `nextjs-kit`, but given only Drupal supports preview at the moment, it's fine to keep it here.

Small typo on the function name:
```suggestion
export default function PreviewRibbon() {
```

starters/next-drupal-starter/components/preview-ribbon.jsx
```diff
@@ -0,0 +1,32 @@
...
+          <button onClick={() => setShow(false)}>
+            <img className="w-8" src="/collapse.svg"></img>
+          </button>
+        </>
+      ) : (
+        <button onClick={() => setShow(true)}>
+          <img className="w-8 rotate-180" src="/collapse.svg"></img>
+        </button>
+      )}
+    </div>
```
CobyPear, line 30: When the ribbon hides, there is some weirdness with the button and the `p` tag stacking on top of each other for a brief moment.

As we discussed in slack, I think we can use `translate-x` for the 'drawer slide' animation.

Here's my current solution, but I couldn't get the collapse button to stay to the left side when closing. It shifts over to the right, then ends up in the correct place, but it would be nice if it stayed in the same place.

Also, I'm not sure `div` is the correct element for the outermost container. Should this be a section? Do we have proper aria labels on everything? in other words, is this component accessible? 

```jsx
    <div className="overflow-hidden">
      <div
        className={`flex justify-between h-14 px-8 py-2 z-10 bg-[#3017A1] text-white transition-all delay-75 w-full ${
          !show ? "translate-x-[calc(100%-5rem)]" : ""
        }`}
      >
        <div
          className={`flex justify-self-center transition-all w-[75vw] ${
            !show && "hidden w-0"
          }`}
        >
          <span className="ml-12 md:ml-20 self-center">Preview Mode Enabled</span>
          <a
            className="border text-black border-black px-4 py-2 mr-4 bg-yellow-300"
            href="/api/clear-preview"
          >
            Exit Preview Mode
          </a>
        </div>
        <button onClick={() => setShow(!show)}>
          <img
            className={`w-8 justify-end ${show ? "rotate-180 ml-8" : ""}`}
            src="/collapse.svg"
          />
        </button>
      </div>
    </div>
```

CobyPear (changes requested):
Really nice work! I think there are some small visual optimizations to be made to the ribbon component, but it is functioning as intended!

CobyPear:
@enZane The `overflow-hidden` was not working for me without adding another div. I also put the button as position absolute, can you take a look? Next was complaining about `Image` and `Link` so I added those too.

CobyPear (approved):
💯

PR #263, opened 2022-09-06
DB-3878: Change Next Starters to use Standalone Output Mode

backlineint:
## What changes were made?

Changed both next starter kits to use standalone output mode, which has a substantial reduction of cold start times.

* Upgraded next to latest version. Since nextjs-kit had next as a direct (and peer?) dependency, this change was made in both next starter kits.
* Rebuilt related package lock files in both next starter kits.
* Updated next config to use standalone output in both next starter kits.
* Updated default build and start commands to use standalone output target.
* Configured monorepo to use default build and start tasks - the standalone build doesn't seem to work in our pnpm/monorepo configuration.

Manually tested builds in both starter kits to confirm that they created a deployable standalone build asset and that it could be served.

## Additional information


packages/nextjs-kit/package.json
```diff
@@ -63,9 +63,9 @@
     "vitest": "^0.18.0"
   },
   "peerDependencies": {
-    "next": "12.1.6",
+    "next": "^12.2.5",
```
CobyPear, line 66: I didn't run into this when setting this package up, but now I am realizing we should be less restrictive here, maybe `>=^12.2.5`. Currently when I try to `npm i` a clone of the drupal starter, it yells at me because the published `nextjs-kit` 1.0.0 wants exactly next 12.1.6.

packages/nextjs-kit/package.json
```diff
@@ -63,9 +63,9 @@
...
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "tailwindcss": "3.1.6"
```
CobyPear, line 69: same with `tailwind`, I think we can support `>=^3.1.6`

CobyPear (changes requested):
Just one small change with the peer dep version, otherwise this seems to be working wonderfully!
I was able to deploy a copy of the Drupal starter to the infra and the initial load seemed a lot faster than normal! Waiting for it to spin down to check the cold-start.

CobyPear (approved):

PR #274, opened 2022-09-12
DB-3654: Block Editor Cover Styles

enZane:
## What changes were made?
- Added support for Cover component
- WordPress-Kit

- Locally
## Additional information

### DIFF
--- a/.changeset/gentle-pumas-compare.md
+++ b/.changeset/gentle-pumas-compare.md
@@ -1,0 +1,5 @@
+---
+"@pantheon-systems/wordpress-kit": minor
+---
+
+Add support for cover component from WordPress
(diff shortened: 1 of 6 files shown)


CobyPear (changes requested):
Looking good! I am noticing when I select the text in the cover and give it some styles like bold and/or italic, the color reverts to black. Not sure if we're meaning to support these text styles in the cover or not. Probably should.

Expected:
image

Actual: 
image

Also, if you insert an image into the text block, the width/height of the image does not seem to be respected. I'm more OK with this one but I wonder what the usecase is and if it is a problem?

Example:
WP
image


Next.js
image

enZane:
I guess the text bug should be covered in this ticket https://getpantheon.atlassian.net/browse/DB-3656 because the fix is the same, but let me know if you want to fix in it

On my way the image stuff 👀

CobyPear (approved):
Awesome!

PR #276, opened 2022-09-13
DB-3781: ContentWithImage component

mitchellmarkoff:
## What changes were made?
Added a `<ContentWithImage />`  component that can be used for a WP Post or a Drupal Article. 

nextjs-kit

Component was implemented locally into the WP and Drupal starters. Unit test written.

## Additional information


packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,86 @@
...
+	date?: string;
+}
+
+/**
+ *
+ * @param props - The props needed for the ContentWithImage component
+ * @param props.title - The title of your content
+ * @param props.featuredImage - A URL for the image source of your content
+ * @param props.content - A string of elements to make up the content of your post
+ * @param props.imgUrl - The image URL coming from your enviornment config
```
CobyPear, line 19: This description can be a bit more specific and maybe could use a link to the docs on `next/image`

CobyPear (changes requested):
See comments and also changeset please!

.changeset/friendly-mugs-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pantheon-systems/nextjs-kit": patch
+---
+
+DB-3781: Added ContentWithImage component to be used for a WP Post or Drupal Article
```
CobyPear, line 5: The changeset messages will be appended directly to the changelog. We don't need the Jira ticket number here.

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,68 @@
...
+	imageProps,
+	content,
+	date,
+	previousPagePath,
+}: ContentProps) => {
+	return (
+		<article className="prose lg:prose-xl mt-10 mx-auto">
+			<h1>{title}</h1>
+			{date ? (
+				<p className="text-sm text-gray-600">{new Date(date).toDateString()}</p>
```
CobyPear, line 40: I'm wondering if we should leave the date formatting up to the user and just take in and spit out the string they give us here. The way it is now a user could pass in an invalid date without realizing it and this would break.
Also I think this is considered a side-effect and we should try to keep these components as pure as possible.

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,68 @@
...
+			) : null}
+
+			{previousPagePath ? (
+				<Link passHref href={previousPagePath}>
+					<a className="font-normal">Back &rarr;</a>
+				</Link>
```
CobyPear, line 46: This is not very scaleable. Let's just make this always go back.
see https://nextjs.org/docs/api-reference/next/router#routerback

packages/nextjs-kit/src/__tests__/snapshotTests/contentWithImage.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+		'<blockquote class="wp-block-quote"><p>a Quote</p><cite>from QA</cite></blockquote>\n' +
+		'\n' +
+		'\n' +
+		'\n' +
+		'<figure class="wp-block-pullquote" style="border-color:#0073a8"><blockquote class="has-text-color has-dark-gray-color"><p>a pull quoteeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee</p><cite>from qa</cite></blockquote></figure>\n' +
+		'\n' +
+		'\n' +
+		'\n' +
+		'<p>An<strong>other block</strong></p>\n',
+	imgUrl: 'https://dev-decoupled-wordpress-qa.pantheonsite.io',
```
CobyPear, line 27: This seems like it should be `wp-content/uploads/2022/08/pizza.jpeg` or something similar, not the domain part of the URL.

packages/nextjs-kit/src/__tests__/snapshotTests/__snapshots__/contentWithImage.test.tsx.snap
```diff
@@ -0,0 +1,100 @@
...
...       srcset="/_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=640&q=75 640w, /_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=750&q=75 750w, /_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=828&q=75 828w, /_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=1080&q=75 1080w, /_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=1200&q=75 1200w, /_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=1920&q=75 1920w, /_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=2048&q=75 2048w, /_next/image?url=https%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.iohttps%3A%2F%2Fdev-decoupled-wordpress-qa.pantheonsite.io&w=3840&q=75 3840w"
```
CobyPear, line 38: Kind of surprised this works without mocking `next/image`; looking at the path it doesn't look like it would resolve correctly.

CobyPear (changes requested):

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * @param props.title - The title of your content
+ * @param props.imageProps - All props the user wishes to pass to the next/image component
+ * @remarks
+ * imageProps is an optional prop to be used if there is an image to be associated with the content.
+ * If imageProps is used it is required that the user passes in values for src, width, and height
+ * See the documentation link below for more information on optional and required props
+ * @see {@link https://nextjs.org/docs/api-reference/next/image} for all next/image documentation
+ * @param props.content - A string of elements to make up the content on the post
+ * @param props.date - An optional date to be displayed on the post
+ * @param props.previousPagePath - The path of the previous page to navigate back to
```
CobyPear, line 25:
```suggestion
```

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,63 @@
...
+/**
+ *
+ * @param props - The props needed for the ContentWithImage component
+ * @param props.title - The title of your content
+ * @param props.imageProps - All props the user wishes to pass to the next/image component
+ * @remarks
+ * imageProps is an optional prop to be used if there is an image to be associated with the content.
+ * If imageProps is used it is required that the user passes in values for src, width, and height
+ * See the documentation link below for more information on optional and required props
+ * @see {@link https://nextjs.org/docs/api-reference/next/image} for all next/image documentation
```
CobyPear, line 22: Let's move the `@remarks` to the bottom of the TSDoc comment, after the `@returns`, and we can move the `@see {@link}` part into the `@param` comment for `props.imageProps`

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+			<a onClick={() => router.back()} className="font-normal">
+				Back &rarr;
+			</a>
+
+			<div className="mt-12 max-w-lg mx-auto lg:grid-cols-3 lg:max-w-screen-lg">
+				{imageProps ? (
+					<div
+						className="relative w-full rounded-lg shadow-lg overflow-hidden mb-10"
+						style={{ height: '50vh' }}
```
CobyPear, line 48: This is another artifact from the quick'n'dirty implementation of this in the starters. When I give the imageProps a height and width, I would expect it to be respected by this container as well.

Code:
```jsx
<ContentWithImage
    title={title}
    content={body.processed}
    imageProps={{ src: imgSrc, width: 200, height: 100 }}
/>
```

Result:
image

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param props - The props needed for the ContentWithImage component
+ * @param props.title - The title of your content
+ * @param props.imageProps - All props the user wishes to pass to the next/image component
+ * @remarks
+ * imageProps is an optional prop to be used if there is an image to be associated with the content.
+ * If imageProps is used it is required that the user passes in values for src, width, and height
+ * See the documentation link below for more information on optional and required props
+ * @see {@link https://nextjs.org/docs/api-reference/next/image} for all next/image documentation
+ * @param props.content - A string of elements to make up the content on the post
```
CobyPear, line 23:
```suggestion
 * @param props.content - Content from a CMS, usually a string of raw HTML. This string will be set as `dangerouslySetInnerHTML`
```

packages/nextjs-kit/src/__tests__/data/examplePostData.json
```diff
@@ -0,0 +1,8 @@
+[
+	{
+		"title": "Example Post with Image",
+		"date": "2022-08-04",
+		"featuredImage": "https://dev-decoupled-wordpress-qa.pantheonsite.i/wp-content/uploads/2022/08/pizza.jpeg",
+		"content": "<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>\n \n \n \n \n \n <blockquote class=\"wp-block-quote\"><p>a Quote</p><cite>from QA</cite></blockquote>\n <figure class=\"wp-block-pullquote\" style=\"border-color:#0073a8\"><blockquote class=\"has-text-color has-dark-gray-color\"><p>a pull quoteeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee</p><cite>from qa</cite></blockquote></figure>\n \n \n \n <p>An<strong>other block</strong></p>\n"
```
CobyPear, line 6: Can we use a more generic post without the 'a pull quoteeeeeeeeeeeeeeeeeeeeeeeee' stuff in it? :)

packages/nextjs-kit/src/__tests__/snapshotTests/contentWithImage.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+import ContentWithImage from '../../components/contentWithImage';
+import examplePostData from '../data/examplePostData.json';
+import { vi } from 'vitest';
+
+vi.mock('next/image', () => ({
+	__esModule: true,
+	default: (props: any) => {
+		return <img {...props} />;
+	},
+}));
```
CobyPear, line 11: It is likely that we will need to mock `next/image` anytime we use `next/image`, can we move this out to `__mocks__` like we have in the `next-drupal-starter`?

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,63 @@
...
+	content,
+	date,
+}: ContentProps): JSX.Element => {
+	const router = useRouter();
+	return (
+		<article className="prose lg:prose-xl mt-10 mx-auto">
+			<h1>{title}</h1>
+			{date ? <p className="text-sm text-gray-600">{date}</p> : null}
+
+			<a onClick={() => router.back()} className="font-normal">
```
CobyPear, line 40:
```suggestion
			<a onClick={() => router.back()} className="font-normal cursor-pointer">
```

CobyPear (changes requested):

packages/nextjs-kit/src/__mocks__/next/image.tsx
```diff
@@ -0,0 +1,4 @@
+export default function image(props: any) {
```
CobyPear, line 1: Looks like there is a way to type this:

```suggestion
export default function image(
	props: React.DetailedHTMLProps<
		React.ImgHTMLAttributes<HTMLImageElement>,
		HTMLImageElement
	>,
) {
```

CobyPear (changes requested):

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+			<a onClick={() => router.back()} className="font-normal cursor-pointer">
+				Back &rarr;
+			</a>
+			<div className="mt-12 max-w-lg mx-auto lg:grid-cols-3 lg:max-w-screen-lg">
+				{imageProps ? (
+					<div className="rounded-lg shadow-lg overflow-hidden mx-auto">
+						<Image {...imageProps} />
+					</div>
+				) : null}
```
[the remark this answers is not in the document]
mitchellmarkoff: Ah I had interpreted that a bit differently so this implementation counts on the user still passing some layout props but you are right, it makes more sense to only allow the user to pass `src` if we are not taking on full styling support now.

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -40,8 +40,17 @@ const ContentWithImage: React.FC<ContentProps> = ({
...
+					<div
+						className="relative w-full rounded-lg shadow-lg overflow-hidden mb-10"
+						style={{ height: '50vh' }}
+					>
+						<Image
+							priority
+							src={imageProps}
+							layout="fill"
+							objectFit="cover"
+							alt={'Featured Image'}
```
CobyPear, line 52: The alt text should be the only other configurable option I think

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import Image from 'next/image';
+import type { ImageProps } from 'next/image';
+import { useRouter } from 'next/router';
+
+interface ContentProps {
+	title: string;
+	imageProps?: ImageProps['src'];
+	content: string;
+	date?: string;
```
CobyPear, line 10: Small thing, but let's put the optional props at the bottom of the interface.
Also we will want to add `imageAltText` as a prop as well, and we should fall back to the `title` if we don't get alt text from the user.

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ *
+ * @param props - The props needed for the ContentWithImage component
+ * @param props.title - The title of your content
+ * @param props.imageProps - All props the user wishes to pass to the next/image component. @see {@link https://nextjs.org/docs/api-reference/next/image} for all next/image documentation
```
CobyPear, line 17: This comment will need to be updated

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * @param props.content - Content from a CMS, usually a string of raw HTML. This string will be set as `dangerouslySetInnerHTML`
+ * @param props.date - An optional date to be displayed on the post
+ * @returns A component with a featured image and content passed by the user
+ * @remarks
+ * imageProps is an optional prop to be used if there is an image to be associated with the content.
+ * If imageProps is used it is required that the user passes in values for src, width, and height.
```
CobyPear, line 23: This will also need to be updated

CobyPear (changes requested):

CobyPear (approved):

PR #279, opened 2022-09-13
DB-3782: Recipe Component

mitchellmarkoff:
## What changes were made?
A `<Recipe />` component was added to the nextjs-kit for use with the Drupal starter.

`nextjs-kit`

Component was locally implemented into the `next-drupal-starter`

## Additional information


packages/nextjs-kit/src/__tests__/snapshotTests/__snapshots__/recipe.test.tsx.snap
```diff
@@ -0,0 +1,157 @@
...
+      class="relative max-w-lg mx-auto min-w-full h-[50vh] rounded-lg shadow-lg overflow-hidden mt-12 mb-10"
+    >
+      <span
+        style="box-sizing: border-box; display: block; overflow: hidden; background: none; opacity: 1; border: 0px; margin: 0px; padding: 0px; position: absolute; top: 0px; left: 0px; bottom: 0px; right: 0px;"
+      >
+        <img
+          data-nimg="fill"
+          decoding="async"
+          sizes="100vw"
+          src="/_next/image?url=https%3A%2F%2Fdev-decoupled-drupal-qa.pantheonsite.io%2Fsites%2Fdefault%2Ffiles%2F2022-07%2Fguacamole.jpeg&w=3840&q=75"
```
CobyPear, line 38: Same as the other PR, surprised this works!

CobyPear (changes requested):

packages/nextjs-kit/src/__tests__/snapshotTests/recipe.test.tsx
```diff
@@ -0,0 +1,45 @@
+import { render } from '@testing-library/react';
+import Recipe from '../../components/recipe';
+
+/**
+ * @vitest-environment jsdom
+ */
+
+const recipe = {
```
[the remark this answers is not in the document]
mitchellmarkoff: Since the data was so small I figured it was fine in the test but I like the idea of keeping everything in the data folder. Will move it in there

packages/nextjs-kit/src/__mocks__/image.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+export default function image(
+	props: React.DetailedHTMLProps<
+		React.ImgHTMLAttributes<HTMLImageElement>,
+		HTMLImageElement
+	>,
+) {
+	return <img {...props} />;
+}
```
CobyPear, line 9: I was refactoring the ContentWithImage component and I don't think this mock is necessary. I'm not sure it is even doing what we are expecting it to because I get the same results with or without the mock.

.changeset/small-ants-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pantheon-systems/nextjs-kit": patch
+---
+
+DB-3782: Add Recipe Component for use in next-drupal-starter
```
CobyPear, line 5: No need for the ticket number here.
Also try not to reference the starters in the packages because on `npm` they are not connected at all.

packages/nextjs-kit/src/components/recipe.tsx
```diff
@@ -0,0 +1,90 @@
...
+ * @param props.category - The identifying category of your recipe
+ * @param props.imageProps - All props the user wishes to pass to the next/image component @see {@link https://nextjs.org/docs/api-reference/next/image} for all next/image documentation
+ * @param props.ingredients - An array of strings with ingredients to be displayed
+ * @param props.instructions - Instructions from a CMS, usually a string of raw HTML. This string will be set as `dangerouslySetInnerHTML`
+ * @returns A recipe component with content and an optional image passed by the user
+ * @remarks
+ * imageProps is an optional prop to be used if there is an image to be associated with the content.
+ * If imageProps is used it is required that the user passes in value for the images src
+ */
+const Recipe: React.FC<RecipeProps> = ({
```
CobyPear, line 28: I thought this was going to get some more refactoring instead of just a copy -> paste from the original component?

packages/nextjs-kit/src/components/recipe.tsx
```diff
@@ -0,0 +1,90 @@
...
+import type { ImageProps } from 'next/image';
+import { useRouter } from 'next/router';
+
+interface RecipeProps {
+	title: string;
+	category: string;
+	imageProps?: ImageProps;
+	ingredients: string[];
+	instructions: string;
+	children?: JSX.Element;
```
CobyPear, line 12: Group optional props to the bottom of the interface

packages/nextjs-kit/src/components/recipe.tsx
```diff
@@ -0,0 +1,90 @@
...
+					<span className="text pb-2 pr-3 text-sm text-slate-400">
+						{category}
+					</span>
+				</div>
+			</header>
+			{imageProps ? (
+				<div className="relative max-w-lg mx-auto min-w-full h-[50vh] rounded-lg shadow-lg overflow-hidden mt-12 mb-10">
+					<Image {...imageProps} />
+				</div>
+			) : null}
```
CobyPear, line 57: Check the refactor on the ContentWithImage component. I think if we are going to keep it as a cover image type thing, we should be consistent with the props we take in.

I think it could even make sense to extract the `featuredImage` component into an atom that can be used in this Recipe component, as well as in ContentWithImage and any future components with a 'featuredImage', but perhaps we should wait on that.

CobyPear (changes requested):

packages/nextjs-kit/src/components/recipe.tsx
mitchellmarkoff: Ah, I had thought that the refactoring to do here was only to add support for different recipe fields which we decided to use `children` for. Aside from the new comments, what additional areas are you thinking should be looked at for refactoring?

CobyPear: Oh yes, that sounds familiar, my bad! 😅

packages/nextjs-kit/src/__tests__/data/exampleRecipeData.json
```diff
@@ -0,0 +1,27 @@
...
+			"salt",
+			"lime",
+			"black pepper",
+			"cilantro"
+		],
+		"field_media_image": {
+			"uri": {
+				"url": "https://dev-decoupled-drupal-qa.pantheonsite.io/sites/default/files/2022-07/guacamole.jpeg"
+			}
+		},
```
CobyPear, line 17: There should be alt text available here as well. See https://github.com/pantheon-systems/decoupled-kit-js/blob/canary/starters/next-drupal-starter/__tests__/data/umamiEnRecipesData.json#L146 for the full shape of the `field_media_image` object.

CobyPear (changes requested):
Unsure of where this is coming from, but there is a regression in the homepage header style when I import and use the new recipe component.

image

mitchellmarkoff:
> Unsure of where this is coming from, but there is a regression in the homepage header style when I import and use the new recipe component.
> 
> image

Still digging to see where this came from but that regression is currently live in `canary` so it is outside of this branch.

CobyPear (approved):
:shipit:

PR #287, opened 2022-09-14
Fix preview

CobyPear:
## What changes were made?
Apparently the prettier nuke broke the `preview.js` file worse that initially thought as we found out at a bad time. We may want to pull this change into Main too so we can get this change mirrored to the starter template, or we can ninja push to that repo but I'd rather not :)

`preview.js` was un-reverted
also this remove the re-entering of pre mode before ci:publish happens to avoid publishing the wrong tag.

tested preview locally which was failing before and is failing on canary.

## Additional information


backlineint (approved):
I don't understand how this particular regression happened as part of the prettier work, but the end result here makes sense to me.

PR #293, opened 2022-09-16
DB-4071: refactor drupal-kit bundle

CobyPear:
## What changes were made?
- Remove rollup.config.js in favor of rollupOptions in vite config
- Use vite for the build step
- Add the vite-dts-plugin and emit d.ts files in the bundle
- Restructure folders to match the other kits
- Add build and watch aliases to root monorepo
- Remove rollup command from wp-kit
- Also added the eslint-prettier-config to `drupal-kit`. I have been getting errors during lint-staged without it. I'm not sure why the other packages don't seem to complain.

`drupal-kit`, one sneaky edit to package.json of `wordpress-kit`

locally

## Additional information

Don't forget to
if needed!


packages/drupal-kit/src/lib/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as defaultFetch } from './defaultFetch';
```
CobyPear, line 1: @backlineint I flattened out the folders a bit, I wasn't sure the reason for `fetch` having its own directory but we can bring that back if needed.

backlineint (approved):

PR #299, opened 2022-09-22
DB-3605: Grid Component

mitchellmarkoff:
## What changes were made?
A Grid component to be used in the nextjs starters was added to the nextjs-kit.

nextjs-kit

The new Grid component was implemented locally and tested with all of our current datasets that get placed into a Grid.
## Additional information

Don't forget to
if needed!

### DIFF
--- a/packages/nextjs-kit/src/components/grid.tsx
+++ b/packages/nextjs-kit/src/components/grid.tsx
@@ -29,7 +29,7 @@ export const Grid = ({
  *
  * @param Component - A component that takes in content that is to be displayed on the grid
  * @remarks
- * The Component used must accept the data to be displayed as 'content' to function properly
+ * The Component used must accept the data to be displayed as `content` to function properly
  * @example
  * ```
  * const ArticleCard= ({ content }) => {
@@ -45,16 +45,16 @@ export const Grid = ({
  * @example
  * ```
  * const MyPage = ({ myArticles }) => {
- * const ArticleGrid = withGrid(ArticleCard)
- * 	return (
- * 		<>
- * 			<ArticleGrid
- * 				data={myArticles}
- * 				cols={4}
- * 				FallbackComponent={<span>No Data Found</span>}
- * 			/>
- * 		</>
- * 	)
+ *  const ArticleGrid = withGrid(ArticleCard)
+ * 	 return (
+ * 		 <>
+ * 			 <ArticleGrid
+ * 				 data={myArticles}
+ * 				 cols={4}
+ * 				 FallbackComponent={<span>No Data Found</span>}
+ * 			 />
+ * 		 </>
+ * 	 )
  * }
  * ```
  */


CobyPear (changes requested):

packages/nextjs-kit/tailwind.config.js
```diff
@@ -1,6 +1,12 @@
 /** @type {import('tailwindcss').Config} */
 module.exports = {
 	content: ['./src/components/**/*.tsx'],
+	safelist: [
+		{
+			pattern: /grid-cols-([0-9])/,
+			variants: ['lg'],
+		},
+	],
```
CobyPear, line 9: I'm skeptical that this is the best solution to the cols problem, the docs say this is a last resort kind of thing. Is there truly no better way?

packages/nextjs-kit/src/components/grid.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * @returns A Higher Order Component that returns the data mapped to the Component in a grid
+ */
+export const withGrid = (Component: React.ElementType) => {
+	const GridedComponent = ({
+		data,
+		FallbackComponent,
+		cols,
+		...props
+	}: {
+		data: Record<string, string | number>[];
```
CobyPear, line 44: Let's use a generic here

CobyPear (changes requested):

packages/nextjs-kit/tailwind.config.js
mitchellmarkoff: I wrestled with this for awhile and have yet to find a solution I think is better. I tried the helper function route outlined [here](https://stackoverflow.com/a/72268889) but that did not fix the problem. The only other way I found to get the dynamic columns working was by side stepping tailwind and adding an inline style like so
```
style={{
	gridTemplateColumns: `repeat(${cols}, minmax(0, 1fr))`,
}}
```
This solution didn't feel any better right and makes it more difficult to add a media query for breakpoint definitions.

Let me know what you think 👍

packages/nextjs-kit/src/__tests__/snapshotTests/__snapshots__/grid.test.tsx.snap
```diff
@@ -0,0 +1,124 @@
...
+      </h1>
+      <p>
+        Fourth item in the grid
+      </p>
+    </div>
+  </div>
+</DocumentFragment>
+`;
+
+exports[`<Grid /> > should render FallbackComponent instead of 'Grid' 1`] = `
```
CobyPear, line 44: Looks like this is rendering the grid?

packages/nextjs-kit/tailwind.config.js
```diff
@@ -1,6 +1,12 @@
 /** @type {import('tailwindcss').Config} */
 module.exports = {
 	content: ['./src/components/**/*.tsx'],
+	safelist: [
+		{
+			pattern: /grid-cols-([0-9])/,
```
CobyPear, line 6: I think we need the prefix here
```suggestion
			pattern: /ps-grid-cols-([0-9])/,
```

CobyPear (changes requested):

packages/nextjs-kit/tailwind.config.js
mitchellmarkoff: Sorry I forgot to add a comment, I tested without the prefix and it seems that it is not necessary.

.changeset/wet-foxes-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pantheon-systems/nextjs-kit': patch
```
CobyPear, line 2: I think this can be a minor too given your reply on #303

PR #301, opened 2022-09-23
DB-4217: Add tailwindcss prefix

CobyPear:
## What changes were made?
- Add the `prefix` option to the tailwindcss config
- Refactor all tailwindcss classes to add the prefix
- Also took this opportunity to refactor the Paginator component a bit and increase the accessibility with proper aria labels, a nav container, and `ul>li` markup. 
- Added the eslint-prettier plugin to the nextjs kit since the lint step complains when I do not have it as a dev dependency.

`nextjs-kit`, updated snapshot tests for the pagination component in `next-drupal`
Locally w/ a screen reader

### Note:
Going forward we will need to use `ps-` for all tailwind styles in the nextjs-kit. We should follow suit for any other libraries that export components. (Will ps- work for all of them, what if you import from 2 libraries with ps-, will it break?)

There is some significant effort to switch the `buttons` to `a` tags which is the better way to do things from my research. I will write something up for this.

I found this article to be very informative: https://www.a11ymatters.com/pattern/pagination/#wrap-the-pagination-links-in-a-nav-element

Don't forget to
if needed!

### DIFF
--- a/.changeset/small-eggs-work.md
+++ b/.changeset/small-eggs-work.md
@@ -1,0 +1,6 @@
+---
+'@pantheon-systems/nextjs-kit': minor
+---
+
+Added the prefix option in the tailwindcss config so the styles generated from
+nextjs-kit will not override default tailwindcss classes in a project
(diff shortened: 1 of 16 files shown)


enZane (approved):
LGTM! Tested changing tokens on starter and that doesn't impact the Next-Kit

PR #304, opened 2022-09-27
DB-4062: Block editor button styles

enZane:
## What changes were made?
- Added button component
- Added getColorByName to the color class

- WordPress-Kit

- With Next-WP-Starter

## Additional information
- Looks like alignment is bugged from WP, doesn't work on WP preview, and doesn't add any class to the produced HTML

Don't forget to
if needed!

### DIFF
--- a/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Buttons.ts
+++ b/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Buttons.ts
@@ -1,0 +1,96 @@
+const justifications = {
+	'&.is-content-justification-left': {
+		justifyContent: 'flex-start',
+	},
+	'&.is-content-justification-center': {
+		justifyContent: 'center',
+	},
+	'&.is-content-justification-right': {
+		justifyContent: 'flex-end',
+	},
+	'&.is-content-justification-space-between': {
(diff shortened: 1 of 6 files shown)


packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Buttons.ts
```diff
@@ -0,0 +1,96 @@
...
+		alignItems: 'center',
+		'justify-content': 'flex-start',
+		...justifications,
+		'max-width': '650px',
+		'&.alignwide': {
+			maxWidth: '850px',
+		},
+		'&.alignfull': {
+			'@media (min-width:720px)': {
+				// sets a negative margin to allow full width tables to span past the
```
CobyPear, line 50:
```suggestion
				// sets a negative margin to allow full width element to span past the
```

CobyPear (approved):
Just one very insignificant change on a comment. LGTM.

packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Buttons.ts
enZane: Oh, I forgot to change it, nice catch! Done :p

PR #308, opened 2022-09-28
DB-3655: Add file block

enZane:
## What changes were made?
- Added File block support
- Added general button styles
- Used those styles on buttons block

- WordPress-Kit

- Locally
## Additional information

Don't forget to
if needed!


CobyPear (approved):
It seems there are not many options to style this one from the WP side unless I'm missing something. The only thing that doesn't seem to work is the 'open in new tab' option, but that is out of scope here.

✅

PR #320, opened 2022-10-03
DB-3186: Add Pagination To Page and Post List

mitchellmarkoff:
## What changes were made?

- Pagination was added to the Page and Post Lists
- Page and Posts queries return 12 results instead of 10

next-wordpress-starter

In a local development environment using a paginated dataset 

## Additional information


backlineint (changes requested):
Noticed a few things that seem like regressions here. Is this the first time we're applying the pager outside of the example page? These may be things that are a result of the way we built the pager. But I think we would have to address these before using the pager elsewhere.

* The grid on /pages and /posts is now smaller. I'd expect the size to stay the same and match the homepage.
* Link styling has changed on cards - text now has underline text decoration.
* Clicking on the pages or posts links in the header redirects you from /pages to /pages/1 I'd expect the link to behave the same. The Drupal pager example seems to accommodate this: https://next-drupal-starter-umami-canary.pantheon.site/examples/pagination but this implementation always redirects.

This one maybe feels like more of a nice to have:
* If there is only one page of results, page 1/1 should not display (the pager itself does hide, I'd expect both of these elements to follow the same rules.)

mitchellmarkoff:
> Noticed a few things that seem like regressions here. Is this the first time we're applying the pager outside of the example page? These may be things that are a result of the way we built the pager. But I think we would have to address these before using the pager elsewhere.
> 
> * The grid on /pages and /posts is now smaller. I'd expect the size to stay the same and match the homepage.
> * Link styling has changed on cards - text now has underline text decoration.
> * Clicking on the pages or posts links in the header redirects you from /pages to /pages/1 I'd expect the link to behave the same. The Drupal pager example seems to accommodate this: https://next-drupal-starter-umami-canary.pantheon.site/examples/pagination but this implementation always redirects.
> 
> This one maybe feels like more of a nice to have:
> 
> * If there is only one page of results, page 1/1 should not display (the pager itself does hide, I'd expect both of these elements to follow the same rules.)

Thanks for the comments here Brain these are good notes. I will drill into these and see if this work can be done from the starter itself or if we will need to make adjustments to the pager component in `nextjs-kit`.

mitchellmarkoff:
> * Clicking on the pages or posts links in the header redirects you from /pages to /pages/1 I'd expect the link to behave the same. The Drupal pager example seems to accommodate this: https://next-drupal-starter-umami-canary.pantheon.site/examples/pagination but this implementation always redirects.

Updated the PR with your feedback. I am not sure I was clear on this point as it seemed the links here are working the same as in the drupal starter. What I did do with this is now the header redirects to /pages/1 immediately. Let me know if I was mistaken on this.

starters/next-wordpress-starter/components/header.jsx
```diff
@@ -7,8 +7,8 @@ export default function Header() {
 				<ul className="flex flex-row flex-wrap sm:flex-nowrap list-none justify-between max-w-screen-sm mx-auto">
 					{[
 						['🏠 Home', '/'],
-						['📰 Posts', '/posts'],
-						['📑 Pages', '/pages'],
+						['📰 Posts', '/posts/1'],
```
backlineint, line 10: Ideally we don't want to hardcode the links here to include `/1`. `/posts` and `/pages` should be valid paths that are not redirected. The Drupal starter example allows `/examples/pagination` and does not redirect to `/examples/pagination/1` but this implementation automatically redirects to `/1` for some reason.

backlineint:
> Updated the PR with your feedback. I am not sure I was clear on this point as it seemed the links here are working the same as in the drupal starter. What I did do with this is now the header redirects to /pages/1 immediately. Let me know if I was mistaken on this.

The difference between the two is that the Drupal example allows `/examples/pagination` and does not redirect that to `/examples/pagination/1` but this implementation does not seem to allow `/pages` and forces a redirect to `/pages/1`. `/pages` would be a cleaner path if we can allow it.

Testing this did uncover a possibly larger problem related to this. Currently the links use relative paths, which carry the page number through to the URL. For example `http://localhost:3000/pages/1/block-editor-testing` That needs to be `http://localhost:3000/pages/block-editor-testing`. Technically right now both of those paths are valid, which is bad for SEO (and would also be even worse if that page bumps to the second page of results)

backlineint (changes requested):
Noticed one other visual issue while testing. The pager lands with essentially no spacing between the bottom row of results. This looks especially bad on a sub-page that has a small amount of results:

Screenshot at Oct 10 10-32-24

We should give that a little more breathing room.

starters/next-wordpress-starter/components/header.jsx
mitchellmarkoff: Ah I see more clearly now, this has to do with with routing being enabled on the pager, I'll disable that feature to honor the `/pages` route.

backlineint (approved):
Looks great @mitchellmarkoff - thanks for working through all of the feedback here.

starters/next-wordpress-starter/lib/Pages.js
```diff
@@ -25,7 +25,7 @@ export async function getAllPagesUri() {
 export async function getLatestPages() {
 	const query = gql`
 		query LatestPagesQuery {
-			pages(first: 10) {
+			pages(first: 12) {
```
backlineint, line 28: @mitchellmarkoff I should have caught this in review. first: 12 in the query here is limiting the results to 12 even in cases where we have substantially more content. Instead we want to get all pages, but only show 12 results per page.

Quickly tried removing first: 12, but then was only seeing 10 results.

PR #322, opened 2022-10-03
DB-3969: Update seo.jsx with new Gatsby Head API

enZane:
## What changes were made?
- Refactor seo.jsx removing Helmet component and transforming it into a flex component to be used on Head functions
- Refactored every page to use that new seo style

- Gatsby-WP-Starter
- Locally, reviewing the HTML generated
## Additional information


.changeset/dull-clouds-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pantheon-systems/gatsby-wordpress-starter': major
```
backlineint, line 2: Since this does not appear to be a breaking change, I think this should be a minor change rather than a major change.

backlineint (changes requested):

PR #328, opened 2022-10-10
DB-3622: Implement `nextjs-kit` components into drupal starter

mitchellmarkoff:
## What changes were made?

- Relevant nextjs-kit were implemented into the next-drupal-starter
- Components that have been made redundant by the nextjs-kit have been removed
- Snapshots updated

`next-drupal-starter`

Starter was ran locally and no differences appearance or functionality were observed.

## Additional information


starters/next-drupal-starter/__tests__/setupFile.js
```diff
@@ -112,6 +112,17 @@ export const restHandlers = [
 	});
 });
 
+process.env = {
+	...process.env,
+	__NEXT_IMAGE_OPTS: {
+		deviceSizes: [320, 420, 768, 1024, 1200],
+		imageSizes: [],
+		domains: ['default', 'umami'],
```
CobyPear, line 120: In our `vite.config.js` we are defining the current PROFILE as a global env variable. I think you could use that here instead of hardcoding 'deafult' and 'umami', you could use the following:

```suggestion
		domains: [process.env.PROFILE],
```

CobyPear: I see all of the tests passing so I think this is good to go!

starters/next-drupal-starter/components/grid.jsx
```diff
@@ -2,46 +2,6 @@ import Link from 'next/link';
...
-			className={`mt-12 grid gap-5 max-w-content mx-auto lg:grid-cols-${String(
-				cols,
-			)} lg:max-w-screen-lg`}
-		>
-			{children}
-		</div>
-	);
-};
-
-const GradientPlaceholder = () => (
```
CobyPear, line 18: Is this implemented somewhere else?

CobyPear (changes requested):
I think the `<GradientPlaceholder />` component got kiboshed by accident

starters/next-drupal-starter/__tests__/setupFile.js
mitchellmarkoff: This worked but just needed to be passed as `PROFILE` instead of `process.env.PROFILE`. Seems like that is consistent with how we are using the `PROFILE` in the rest of the file.

starters/next-drupal-starter/components/grid.jsx
mitchellmarkoff: Removed by accident, added back 👍

CobyPear (approved):
:shipit:

PR #330, opened 2022-10-11
DB-3186: Update query parameters

mitchellmarkoff:
## What changes were made?
Updates maximum number of posts/pages returned.

`next-wordpress-starter`

Ran starter locally and observes all posts and pages are returned.
## Additional information

WP and graph ql default to displaying 10 results per query when an amount is not specified. There is no good way to return 'all data' other than assigning the `first` query param a value. I adjusted our page and post queries to accept a variable specifying the number of items to be returned. This should get us the desired functionality we want and also allow us to reuse queries a bit more. For example with this set up the home page showing only 12 results can use the same query of the posts list which has all posts. 

### DIFF
--- a/.changeset/late-dryers-collect.md
+++ b/.changeset/late-dryers-collect.md
@@ -1,0 +1,5 @@
+---
+"@pantheon-systems/next-wordpress-starter": patch
+---
+
+Update query parameters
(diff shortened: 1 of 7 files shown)


backlineint (approved):

backlineint:
> There is no good way to return 'all data' other than assigning the first query param a value

This could be a nice feature to add to wordpress-kit, similar to what we added to Drupal State.

PR #340, opened 2022-10-13
DB-2834: add post-install script

enZane:
## What changes were made?
- Added a post install script on the monorepo root that installs husky if is not installed
- Modified the package json to use that script
- decoupled-kit

- Installing and uninstalling the husky hooks
## Additional information
- I'm not sure what to put on the changeset


CobyPear (changes requested):
Only request is to put the script inside of a `scripts` dir.

CobyPear (approved):
🐕

PR #347, opened 2022-10-14
DB-3768: attempt to add the trigger-e2e workflow prior to canary split

backlineint:
## What changes were made?

Attempting to copy the necessary github action to add to the starter repo before the canary split. Won't be 100% positive that this will work until we try running it.

The hacky-er alternative is to instead create a new workflow in decoupled-kit-js that depends on canary-sites-split, and sleeps for 10 min before triggering the Circle CI sdk-test job in qa-automation. Gross, but I think it would work.

Somewhere in between those would be to git commit the workflow during CI. My hope is that this PR has the same end result.

They haven't been. Couldn't think of a practical way to test this locally.

## Additional information

### DIFF
--- a/.github/templates/trigger-e2e.yml.template
+++ b/.github/templates/trigger-e2e.yml.template
@@ -1,0 +1,16 @@
+name: 'Trigger E2E tests on deployment'
+
+on: deployment
+
+jobs:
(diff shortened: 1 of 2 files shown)


CobyPear (approved):
Let's give it a shot!

PR #351, opened 2022-10-17
DB-2625: [next-drupal-starter] Document common issues

CobyPear:
## What changes were made?
- Add a Troubleshooting Local Development doc

docs

## Additional information

### DIFF
--- a/web/docs/Frontend Starters/Next.js/Next.js + Drupal/troubleshooting.md
+++ b/web/docs/Frontend Starters/Next.js/Next.js + Drupal/troubleshooting.md
@@ -1,13 +1,13 @@
 ---
 id: 'next-drupal-troubleshooting'
-title: 'Troubleshooting Local Development'
+title: 'Troubleshooting' 
 sidebar_position: 5
 slug: '/Frontend Starters/Next.js/Next.js + Drupal/Troubleshooting Local Development'
 ---
 
 ## Before You Begin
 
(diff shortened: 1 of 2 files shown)


backlineint (changes requested):
A little hard to predict, but I could see cases where troubleshooting might be more general, rather than related to local development. A specific type of query or content integration that people often get wrong for example. The last portion of the images entry kind of falls into this category as well.

What do you think about maybe just renaming this section 'Troubleshooting'?

CobyPear:
> What do you think about maybe just renaming this section 'Troubleshooting'?

I think that is reasonable. I was being a bit too careful about keeping this doc away from any platform considerations, but I see what you mean.

web/docs/Frontend Starters/Next.js/Next.js + Drupal/troubleshooting.md
```diff
@@ -0,0 +1,46 @@
+---
+id: 'next-drupal-troubleshooting'
+title: 'Troubleshooting' 
+sidebar_position: 5
+slug: '/Frontend Starters/Next.js/Next.js + Drupal/Troubleshooting Local Development'
```
backlineint: The slug should probably change here now that this is just 'Troubleshooting'

backlineint (approved):
I pushed up a minor tweak for the slug here.

PR #358, opened 2022-10-24

enZane:
## What changes were made?
- On Next-Kit:
  - Changed the paddings and margins on the contentWithImage component to allow full page
  - Added ps-wp-content class to wrap base utility classes
- On WordPress-Kit:
  - Changed every component with a shared alignment option to honor full width
  - Added base styles to headings and paragraph

- WordPress-Kit
- Next-Kit

- With Hello World post that has every block component supported

## Additional information


.changeset/tender-students-occur.md
```diff
@@ -0,0 +1,9 @@
+---
+'@pantheon-systems/nextjs-kit': minor
+'@pantheon-systems/wordpress-kit': minor
+---
+
+Modified the contentWithImage component on Next-Kit to remove the paddings and
+allow the content full page. Modified every component on the Wordpress-Kit
+tailwind plugin to respect the full width options. Added base utilities on
+tailwind plugin to modify headings and paragraphs width.
```
CobyPear, line 9: Can we split this into 2 changests, one for each `nextjs-kit` and `wordpress-kit` so the generated changelog is clear for each package?

packages/wordpress-kit/src/lib/tailwindcssPlugin/index.ts
```diff
@@ -13,41 +13,33 @@ import {
...
 
-	const audio = AudioComponent({
-		alignFull: { minWidth: theme('screen.xl', '1280px') },
-	});
+	const audio = AudioComponent();
 
-	const cover = CoverComponent({
-		alignFull: { minWidth: theme('screen.xl', '1280px') },
-	});
+	const cover = CoverComponent();
```
CobyPear, line 42: This looks much cleaner, nice!

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -61,7 +61,7 @@ const ContentWithImage: React.FC<ContentProps> = ({
 			</div>
 
 			<div
-				className="ps-break-words ps-mt-12"
+				className="ps-break-words ps-mt-12 ps-wp-content"
```
CobyPear, line 64: It's a bit strange to use a WordPress specific class in our `nextjs-kit`, as this component is also used in the Drupal starter.

CobyPear (changes requested):

packages/nextjs-kit/src/components/contentWithImage.tsx
CobyPear: I guess I'm not certain the class should go on this component, but instead should go on the implementation in the WordPress based starters, because this class won't do anything for the Drupal starter, right?

packages/wordpress-kit/src/lib/tailwindcssPlugin/index.ts
```diff
@@ -13,41 +13,33 @@ import {
...
 
 /**
  * Tailwindcss plugin that maps WordPress block editor styles to tailwindcss classes.
  */
 export default plugin(function ({ addUtilities, theme, addComponents }) {
 	const color = new ColorUtilities(theme);
 	const font = new FontsUtilities(theme);
+	const base = new BaseUtilities();
+
+	console.log(base.baseUtilities);
```
CobyPear, line 26: oh, just saw this one :)

packages/nextjs-kit/src/components/contentWithImage.tsx
enZane: Should do I add a class name prop for it?

CobyPear: I think instead of adding the class to the component in `nextjs-kit` here, we should add it to the implementation of the components in the `next-wordpress-starter` so that the class doesn't show up in the Drupal implementations.

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -11,6 +11,7 @@ interface ContentProps {
 		src: ImageProps['src'];
 		alt?: ImageProps['alt'];
 	};
+	contentClassName?: string;
```
CobyPear, line 14: Does it have to be a prop on the component? Or can we just add the class to our component straight up?

enZane: In react we need to do this, I'm not the biggest fan of it, but is needed to pass external classes as props

CobyPear: ah I see. It doesn't seem to be showing up on the in the snapshot.

enZane: It won't appear in the snapshot because the join doesn't add empty spaces on undefined values

CobyPear: Oh yeah meakes sense on the snapshot in `nextjs-kit`, I meant the snapshots in the `next-wordpress-starter`, is that expected?

Also, the more I think about this the more I'm ok with it as it adds a way to style the content with custom classes a bit easier.

packages/nextjs-kit/src/components/contentWithImage.tsx
```diff
@@ -61,7 +64,7 @@ const ContentWithImage: React.FC<ContentProps> = ({
 			</div>
 
 			<div
-				className="ps-break-words ps-mt-12 ps-wp-content"
+				className={['ps-break-words ps-mt-12', contentClassName].join(' ')}
```
CobyPear, line 67: Can we also inject it like so to avoid the `.join`?:

```suggestion
                                className={`ps-break-words ps-mt-12 ${contentClassName}`}
```

enZane: The problem with that is that the rendered class when you don't pass a class name will have this `ps-break-words ps-mt-12 ` with the empty space and is not clean at all

What's usually done for this on the tailwind ecosystem is add a function to hide the join, I didn't add it to avoid premature abstraction let me know if that sounds better

packages/nextjs-kit/src/components/contentWithImage.tsx
enZane: The changes on snapshots are here https://github.com/pantheon-systems/decoupled-kit-js/pull/358/files#diff-b8b0b846e64924ebb479741da6e4d519b05d538bf9cee22e7114495d8a9065b8R223 👀

packages/nextjs-kit/src/components/contentWithImage.tsx
enZane: I just did it, because forgot to add a filter to remove the last injected class name if is undefined

CobyPear: Ah I see, makes sense. Thanks!

CobyPear (approved):

PR #361, opened 2022-10-26
DB-1526: Add Pagination to Page and Post Routes

mitchellmarkoff:
## What changes were made?
- Pagination was added to the Page and Post routes of the `gatsby-wordpress-starter`

`gatsby-wordpress-starter`

Changes have been tested locally.

## Additional information

- A small bug in `<Paginator />` was found while working on this, with some of the navigation functionality breaking if there was a hard refresh while using the component. This is because the starter would add a `/` to the url after a manual refresh which was not accounted for in building the pagination button routes. A fix for that is in this PR as it was necessary to support routing.
- Going off of [this](https://github.com/pantheon-systems/decoupled-kit-js/pull/320#pullrequestreview-1131481542) conversation from the previous pagination work, urls at both `/pages` and `/pages/1` are considered valid and supported here with routing.


starters/gatsby-wordpress-starter/__tests__/data/pageTemplateData.json
```diff
@@ -0,0 +1,26 @@
+[
```
mitchellmarkoff, line 1: The `pagesIndex` and  singular `page` now get their data in different structures from one another in separate queries.

starters/gatsby-wordpress-starter/src/templates/postsIndex.jsx
```diff
@@ -46,12 +46,8 @@ export function Head() {
 }
 
 export const pageQuery = graphql`
-	query PostList($offset: Int!, $postsPerPage: Int!) {
-		allWpPost(
-			sort: { fields: [date], order: DESC }
-			limit: $postsPerPage
-			skip: $offset
-		) {
+	query PostList {
```
CobyPear, line 49: We are already querying all posts at build time in `gatsby-node.js`. You should be able to pass in posts via pageContext down to the postsIndex and pageIndex pages.

In other words, we don't need a pageQuery here at all (even though it's been here awhile I think it can be refactored so we can remove this), similar to how there wasn't one in the pagesIndex.

starters/gatsby-wordpress-starter/src/templates/pagesIndex.jsx
```diff
@@ -24,3 +43,24 @@ export default PageIndexTemplate
 export function Head() {
 	return <Seo title="All pages" />
 }
+
+export const pageQuery = graphql`
```
CobyPear, line 47: Since we are fetching all the pages at build time in `gatsby-node.js`, I don't think we require a page query here.

starters/gatsby-wordpress-starter/gatsby-node.js
```diff
@@ -109,48 +109,84 @@ async function createPostArchive({ posts, gatsbyUtilities }) {
...
+
+				component: path.resolve(`./src/templates/pagesIndex.jsx`),
+
+				context: {
+					itemsPerPage,
+					routing,
+				},
+			})
+		}),
+	)
```
CobyPear, line 189: can this be a bit more DRY?

starters/gatsby-wordpress-starter/__tests__/data/pages.json
```diff
@@ -1,26 +1,18 @@
...
-			"uri": "/test-page/"
-		},
-		"page": {
-			"id": "cG9zdDoy",
-			"title": "Sample Page",
-			"uri": "/sample-page/",
-			"featuredImage": null
-		},
-		"next": null
+{
```
CobyPear, line 1: This change seems strange to me - the data is not the same and also it's not in the same shape.

starters/gatsby-wordpress-starter/__tests__/snapshotTests/page.test.jsx
```diff
@@ -2,11 +2,11 @@ import * as Gatsby from 'gatsby'
...
-import pages from '../data/pages.json'
+import data from '../data/pageTemplateData.json'
 
 const StaticQuery = vi.spyOn(Gatsby, 'StaticQuery')
-const page = pages[0]
-const next = pages[1]
+const page = data[0]
+const next = data[1]
 const previous = null
 describe('<PostTemplate />', () => {
```
CobyPear, line 11: oops, this should be `<PageTemplate />` my bad!

starters/gatsby-wordpress-starter/__tests__/snapshotTests/__snapshots__/pagesIndex.test.jsx.snap
```diff
@@ -68,7 +68,70 @@ exports[`<PageIndexTemplate /> > should render with pages 1`] = `
...
       </header>
-      <section />
+      <div
+        class="max-w-screen-lg mx-auto"
+      >
+        <section>
+          <div
+            class="max-w-full"
+          >
+            <h3
```
CobyPear, line 78: I must've missed this on the paginator PR, but we shouldn't use an h3 here as it is not semantic. This should probably be a `span`

CobyPear (changes requested):
Some comments, also I am seeing some images fail to load on the /pages route:
image

starters/gatsby-wordpress-starter/gatsby-node.js
```diff
@@ -112,81 +113,49 @@ async function createPostArchive({ posts, gatsbyUtilities }) {
...
-					context: {
-						itemsPerPage,
-						routing,
-						breakpoints: { start: 4, end: 6, add: 2 },
-					},
-				})
-			}
+	const getPagePath = page => {
+		return page === 0 ? '/posts' : `/posts/${page}`
+	}
```
CobyPear, line 120: You can make this reusable in such a way that you don't have to write it twice (line 142)

starters/gatsby-wordpress-starter/__tests__/snapshotTests/postsIndex.test.jsx
```diff
@@ -8,23 +8,17 @@ const StaticQuery = vi.spyOn(Gatsby, 'StaticQuery')
 
 describe('<PostIndexTemplate />', () => {
 	beforeEach(() => {
-		StaticQuery.mockImplementation(() => {
```
CobyPear, line 11: Same here, why no mock?

starters/gatsby-wordpress-starter/__tests__/snapshotTests/pagesIndex.test.jsx
```diff
@@ -2,28 +2,20 @@ import * as Gatsby from 'gatsby'
...
 
 import PageIndexTemplate from '../../src/templates/pagesIndex'
-import data from '../data/pages.json'
-
-const StaticQuery = vi.spyOn(Gatsby, 'StaticQuery')
+import pages from '../data/pages.json'
 
 describe('<PageIndexTemplate />', () => {
 	beforeEach(() => {
-		StaticQuery.mockImplementationOnce(() => {
```
CobyPear: Curios as to why we no longer need the StaticQuery here?

CobyPear (changes requested):

starters/gatsby-wordpress-starter/gatsby-node.js
```diff
@@ -6,7 +6,7 @@ exports.createPages = async gatsbyUtilities => {
 	const pagPosts = await paginationPostsQuery()
 	const pages = await getPages(gatsbyUtilities)
 	const posts = await getPosts(gatsbyUtilities)
-	const routing = true
+	const routing = false
```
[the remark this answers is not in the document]
mitchellmarkoff: Ah that was outdated from testing, should be true.

starters/gatsby-wordpress-starter/__tests__/snapshotTests/pagesIndex.test.jsx
mitchellmarkoff: The mock is not needed because we are no longer using  a `pageQuery` in the component and instead pass data via `pageContext` from `gatsby-node.js`.

starters/gatsby-wordpress-starter/__tests__/snapshotTests/postsIndex.test.jsx
mitchellmarkoff: Same reason at the other one (:

starters/gatsby-wordpress-starter/__tests__/snapshotTests/pagesIndex.test.jsx
CobyPear: My understanding of these mocks is they are mocking the queries in gatsby-node > `createPages`

mitchellmarkoff: Ah, I could have misunderstood their purpose but I only seen that kind of mock used in their [docs](https://www.gatsbyjs.com/docs/how-to/testing/testing-components-with-graphql/) when testing page queries. The mocks were necessary before but the tests do pass fine without them since I removed the queries from the components themselves. I can add them back if we want but I'm not sure what they are doing without the use of page query.

CobyPear: hmm well maybe we don't need them anywhere then? Are you able to remove it from it from `pagination.test.jsx` as well?

mitchellmarkoff: Yes and done! I removed the mock from `pagination.test.jsx` in this most recent update. We still need mocks in the individual `page` and `post` tests since those are using page queries to get a unique post based off of an id.

starters/gatsby-wordpress-starter/__tests__/snapshotTests/postsIndex.test.jsx
```diff
@@ -1,27 +1,24 @@
...
 		const { asFragment } = render(
 			<PostIndexTemplate
-				data={data}
+				location={location}
 				pageContext={{
-					nextPagePath: data.allWpPost.nodes[1].uri,
-					previousPagePath: null,
+					itemsPerPage: 2,
+					routing: true,
+					posts: [{ post: data.allWpPost.nodes[0] }],
```
CobyPear, line 21: Any reason not to pass in all of the posts here?
```suggestion
					posts: data.allWpPost.nodes,
```

starters/gatsby-wordpress-starter/__tests__/snapshotTests/homepage.test.jsx
```diff
@@ -1,25 +1,20 @@
...
-
 	it('should render with posts', () => {
 		const { asFragment } = render(
 			<Index
-				data={data}
-				pageContext={{ nextPagePath: null, previousPagePath: null }}
+				pageContext={{
+					nextPagePath: null,
+					previousPagePath: null,
+					posts: [
```
CobyPear, line 13: Same here, any reason we can't pass in the array of posts?

starters/gatsby-wordpress-starter/__tests__/snapshotTests/postsIndex.test.jsx
mitchellmarkoff: No reason, and I can add the second post in the example data if we want. The reason it is like this has to do with how the test data is structured and what the component is expecting. The `<PostIndexTemplate />` and `<Post />` are now passed data in different structures.  Data for `<PostIndexTemplate />` is now coming from `gatsby-node` which has a GraphQL alias `post`, this syntax is there to replicate the structure but not change the actual test data. Since the test data is not changing the same file can be used for the `<Post />` test.

mitchellmarkoff: Short answer - We can pass in a whole array of posts but the test data doesn't have the `post` alias that is expected. I could add it in the test data but then we would need two separate data files for the `<Post />` and `<PostIndexTemplate />` tests

CobyPear: If the data structure is different, it should have its own json file instead of reformatting the other test data imo.

mitchellmarkoff: Ah I got mixed up in writing the tests, the same dataset should be fine for each. The data they receive is in the same shape with the `post` alias, they just destructure it differently. Updated the data set with the alias but they should be good to share 👍

CobyPear (approved):

starters/gatsby-wordpress-starter/__tests__/snapshotTests/postsIndex.test.jsx
CobyPear: Right on, thanks!

PR #362, opened 2022-10-26
Revert "DB-4060: Block Editor Alignment Options"

backlineint:
Reverts pantheon-systems/decoupled-kit-js#358

PR #371, opened 2022-10-26
WIP Revert "Revert "DB-4060: Block Editor Alignment Options""

backlineint:
Reverts pantheon-systems/decoupled-kit-js#362

@enZane This un-reverts the Block Editor alignment work. I'm assuming this will need additional refinements / a new changeset.


CobyPear (changes requested):
I think one of the changesets needs to be modified, not sure how much it matters.

We _do_ need to enter `pre` mode for changesets tho!

CobyPear (approved):

PR #374, opened 2022-10-31
DB-4593: Add with-commit script, refactor some GH Actions

CobyPear:
## What changes were made?
- Split the release job into a reusable workflow
- Added a `shared` dir for shared GH Action workflows
- Added a 'with-commit' workflow that should allow us to commit changes in CI like changes to API reference docs
- Upgraded the `checkout` action to v3

CI
## Additional information
Also began thinking of how to use this pattern to exit and enter changeset prerelease mode via CI. I think I have a solution that I will add to another branch.


.github/scripts/with-commit.sh
```diff
@@ -0,0 +1,4 @@
+git config --local user.name "GitHub Action"
```
backlineint, line 1: Maybe change this to 'Pantheon Automation' / bot@getpantheon.com which other projects within pantheon use for this: https://github.com/pantheon-upstreams/wordpress-composer-managed/commits/main 
https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/devops/scripts/deploy-public-upstream.sh#L51-L54

Funky things in the Github UI can happen when commits are made by a real email that we don't own.

backlineint (changes requested):

backlineint (approved):

PR #378, opened 2022-11-02

enZane:
## What changes were made?
- Adds media queries to follow the actual implementation of the layout

- WP-Kit

Locally, reviewed contrasting against Drupal

## Additional information

### DIFF
--- a/.changeset/brown-eels-tap.md
+++ b/.changeset/brown-eels-tap.md
@@ -1,0 +1,6 @@
+---
+'@pantheon-systems/wordpress-kit': minor
+---
+
+Adds media queries to every component and base styles to follow actual layout
+implementation
--- a/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/shared/Alignment.ts
+++ b/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/shared/Alignment.ts
@@ -1,5 +1,13 @@
 export const alignment = {
-	maxWidth: '720px',
+	'@media (min-width: 640px)': {
+		maxWidth: '640px',
+	},
+	'@media (min-width: 768px)': {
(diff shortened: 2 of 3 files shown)


CobyPear (approved):

backlineint:
I'm not positive which PR this belongs on, but guessing this one :)

In this WordPress Post, the header section doesn't align the way I'd expect:

Screenshot at Nov 02 17-34-39

We probably didn't discuss this specifically, but I'd expect that to align with either the default (none), wide, or full widths. Comparing with WordPress, my guess is that this would align to wide.

Screenshot at Nov 02 17-35-56

The header presents a similar issue, but I'd say that is out of scope here.

backlineint:
A little more subtle, but the header and the image aren't quite aligned.

Screenshot at Nov 02 17-37-09

The same problem looks to be happening in the Drupal starter as well.

Related to the previous comment, I'd expect the image to align to wide as well.

backlineint (changes requested):
Added comments in the overview with screenshots.

CobyPear (commented):
I am not seeing the widths applied to the Header block. Is there a reason this block is not included? 
Everything else seems G2G
image

enZane:
> I am not seeing the widths applied to the Header block. Is there a reason this block is not included? Everything else seems G2G image

I saw it before and was because on the WP Block editor, that block has custom margins and that moves it, I guess we should honor custom settings like this

enZane:
> I am not seeing the widths applied to the Header block. Is there a reason this block is not included? Everything else seems G2G image

I saw it before and was because on the WP Block editor, that block has custom margins and that moves it, I guess we should honor custom settings like this

CobyPear:
@enZane I see it now, didn't notice the margin before.

PR #379, opened 2022-11-02

enZane:
## What changes were made?
- Refactor content with image component to have a default styling for the CMS content
- Wrapped the header in a section to apply width logic

- Nextjs-Kit

Locally, tested on Drupal and WordPress flavors, against the Canary site

## Additional information

### DIFF
--- a/.changeset/four-wolves-chew.md
+++ b/.changeset/four-wolves-chew.md
@@ -1,0 +1,5 @@
+---
+'@pantheon-systems/nextjs-kit': minor
+---
+
+Refactor content with image component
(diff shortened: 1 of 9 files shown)


backlineint:
In this case, I'm not sure why these were split out into two PRs. Makes it harder to test and we also can't really merge theme individually.

CobyPear:
> In this case, I'm not sure why these were split out into two PRs. Makes it harder to test and we also can't really merge theme individually.

Sorry, I thought this would make things easier! We can change the target of this PR to the other if that makes things easier.

backlineint (approved):

PR #380, opened 2022-11-03
DB-4065: Block separator styles

mitchellmarkoff:
## What changes were made?

- Added Separator component

`wordpress-kit`

Tested this component locally through a post in the `next-wordpress-starter` that used the separator block. Observed expected styles and behavior.

## Additional information

### DIFF
--- a/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Separator.ts
+++ b/packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Separator.ts
@@ -29,6 +29,18 @@ export const SeparatorComponent = () => {
 				paddingLeft: '2em',
 				fontFamily: 'serif',
 			},
+			'&.has-background': {
+				padding: '0',
+			},
+			'&.has-background:not(.is-style-dots)': {
+				borderBottom: 'none',
+				height: '1px',
+				padding: '0',
+			},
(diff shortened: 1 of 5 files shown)


packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Separator.ts
```diff
@@ -0,0 +1,34 @@
...
+			border: 'none',
+			borderBottom: '2px solid',
+			marginLeft: 'auto',
+			marginRight: 'auto',
+			borderColor: 'black !important',
+			'&.is-style-default': {
+				width: '100px',
+			},
+			'&.is-style-wide': {
+				maxWidth: '650px',
```
enZane, line 13: Looks like the wide separator match the None width on WP, that in the starter is 720px

enZane (changes requested):

packages/wordpress-kit/src/lib/tailwindcssPlugin/components/Separator.ts
enZane, line 13:
```suggestion
				maxWidth: '720px',
```

enZane:
image


The colored separator looks pretty weird, looks like is because the has-background class adds some padding not desired here, and the border is still black

My suggestion is to apply padding: 0 and border-color: currentColor when the block has background color

enZane (approved):

PR #382, opened 2022-11-07
DB-4463: [docs] How to Use Default (non-composer) WP Upstream with Starter Kits

backlineint:
## What changes were made?

* Updated next-wordpress-starter to fail gracefully if used with a WordPress site that does not have menu data.
* Updated gatsby-wordpress-starter to fail gracefully if used with a WordPress site that does not have menu data.
* Refactored next-wordpress-starter to allow for pages and posts with multiple URL arguments.
* Added 'configuring an existing WordPress project' docs entry.

## Additional information


.changeset/eight-needles-lay.md
```diff
@@ -0,0 +1,11 @@
+---
```
CobyPear, line 1: I don't think `web` needs a changeset, or at least we haven't been using them there yet.

Also, to keep the changelog a bit more clean, can the starters have their own changeset?

starters/next-wordpress-starter/lib/Menus.js
```diff
@@ -18,11 +18,19 @@ export async function getFooterMenu() {
...
 
-	return edges.map(({ node }) => node);
+		return edges.map(({ node }) => node);
+	} catch (error) {
+		console.error(
+			'No footer menu data available. Try customizing your query.',
+			error,
+		);
+		return [];
+	}
```
CobyPear, line 35: I think the error handling could be a bit different as to not spam the console with the stack trace:
Should we link to the new doc you made for this or should we wait until we have a stable URL for our docs site?

```suggestion
	try {
		const { menu } = await client.request(query);
		if (!menu) {
			throw new Error('No footer menu data available. Try customizing your query.')
		}

		return menu?.menuItems?.edges?.map(({ node }) => node);		
	} catch (error) {
		console.error(error.message)
		return [];
	}
```

CobyPear (changes requested):
Mostly picky stuff.

I didn't find any regressions and I was able to use a default WP upstream, follow the directions, and use that backend without errors!

starters/next-wordpress-starter/lib/Menus.js
backlineint: Yeah, that is much more pleasant :) And I think we should wait to link to the docs site until it has a final domain.

CobyPear (approved):
LGTM!

PR #383, opened 2022-11-08
DB-4063: Block editor column styles

enZane:
## What changes were made?
- Added columns layout component to the Tailwind plugin

- WordPress-Kit

- Locally
## Additional information

### DIFF
--- a/.changeset/wicked-dogs-report.md
+++ b/.changeset/wicked-dogs-report.md
@@ -1,0 +1,5 @@
+---
+'@pantheon-systems/wordpress-kit': minor
+---
+
+Adds Columns layout component to tailwind plugin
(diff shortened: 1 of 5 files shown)


CobyPear (changes requested):
I think there should be at a small gap between columns or the default paragraph block can look a bit smooshed, maybe add `gap: 1rem;` to the wp-block-columns`? 
image

enZane:
> I think there should be at a small gap between columns or the default paragraph block can look a bit smooshed, maybe add `gap: 1rem;` to the wp-block-columns`? image

Nice catch! I added 1.5rem because 1rem doesn't feel enough :p

CobyPear (approved):

PR #387, opened 2022-11-09
DB-4065: Block separator style adjustment

mitchellmarkoff:
## What changes were made?
- Allow default separator styles to not require a defined class.

`wordpress-kit`

Tested locally with steps that would have previously produced a visual regression.

## Additional information
An issue was found through QA that showed the default style classes for separators are not applied until explicitly selected.
Ex: When a separator is first created, it will have `class="wp-block-separator"` and not be attributed the `is-default-style
` class.
This PR automatically applies default styles when a class is not present and no other styles are selected.


enZane (approved):

packages/wordpress-kit/src/lib/tailwindcssPlugin/Config.ts
```diff
@@ -55,7 +55,6 @@ export const mergeToConfig: Config = {
 		'.wp-block-video',
 		'.wp-block-spacer',
 		'.wp-block-separator',
-		'.wp-block-columns',
```
backlineint, line 58: @mitchellmarkoff Could be missing some context here, but I think this change would essentially be the equivalent of reverting https://github.com/pantheon-systems/decoupled-kit-js/pull/383 - so that would be a regression.

backlineint (dismissed):
Testing this branch I'm seeing all separators with the default hr style for some reason.

backlineint:
> Testing this branch I'm seeing all separators with the default hr style for some reason.

My fault - I was running in dev mode and hadn't run a build locally in order to incorporate the new styles.

packages/wordpress-kit/src/lib/tailwindcssPlugin/Config.ts
mitchellmarkoff: Ah my bad I should have seen this slipped in when I rebased with canary. Not intentional, will get it back in now.

CobyPear (approved):
:shipit:

PR #401, opened 2022-11-18
DB-2874: Create Preview Error Route

mitchellmarkoff:
## What changes were made?
For invalid requests to the preview API endpoint, the user is redirected to a page with one of the following errors:

- The path does not match.
- The requested preview path does not exist.
- Unable to authorize for preview

`next-drupal-starter`

All three types of preview api error were simulated locally and observed to navigate to the correct page with it's expected error message.

## Additional information


CobyPear (changes requested):

starters/next-drupal-starter/pages/preview-error/[[...alias]].jsx
```diff
@@ -0,0 +1,43 @@
...
+} from '../../lib/stores';
+
+export default function PreviewError({ footerMenu, preview }) {
+	const { query } = useRouter();
+	return (
+		<Layout footerMenu={footerMenu} preview={true}>
+			<div className="flex flex-col mx-auto prose-xl mt-20 w-1/2">
+				<h2 className="text-center">
+					🛑 {query.error ? query.error : 'There was an error on the server'} 🛑
+				</h2>
```
Comment (CobyPear) [inline starters/next-drupal-starter/pages/preview-error/[[...alias]].jsx:17]: I think we should put the details in a message query string and put that in a `p` tag for readability. The error and message can be destructured from `useRouter` as well.
```suggestion
				<h2 className="text-center">
					🛑 {error ? error : 'There was an error on the server'} 🛑
				</h2>
				<p>{message}</p>
```

CobyPear (changes requested):
When I click 'clear preview' after an error, it redirects me to the route I was on (`/preview-error`). I think there needs to be a change to the `clear-preview` api route. The following was working to redirect to the homepage but I wonder if there is a more elegant solution that isn't hardcoded? @enZane, any suggestions?

```js
let redirect = new URL(req.headers.referer).pathname;
		redirect = redirect === '/preview-error' ? '/' : redirect;
```

CobyPear (approved):

PR #421, opened 2022-11-29
Tweaking the meta

joshkoenig:
Adding some global keywords and a little tweak to title/description. Next step would be to find a nice image and do some social platform specific tweaking, but I ran out of time.


CobyPear (approved):
Thanks for the contribution @joshkoenig! I will look into adding the social card preview.

PR #425, opened 2022-11-30
DB-4848: getPreview throws unhandled 403 when auth credentials are incorrect

enZane:
## What changes were made?
- Added redirects on error from getPreview

- Next-Drupal-Starter

- Locally commenting the Client_Id on env
## Additional information


CobyPear (changes requested):
Looking good! Looks like the `previewParams.error` check needs to be added to `/pages/[...alias].jsx`– the catchall route– and it should be good to go.

CobyPear (approved):
Approved but holding off on the merge until #428 is in which hopefully fixes the CI issue.

PR #434, opened 2022-12-05
DB-4830: Set Pantheon Environment Env Variables 

mitchellmarkoff:
## What changes were made?
- A PANTHEON_ENVIRONMENT envar is set which is the value of the prefix before -[site-name].appa.pantheon.site in PANTHEON_ENVIRONMENT_URL
- A  IS_LIVE_ENVIRONMENT variable is set to true if PANTHEON_ENVIRONMENT is live and false otherwise
- New environment variables are added to the developer docs
- An example of how to key off of these new variables to source data from a specific BE environment is added to the developer docs

`next-wordpress-starter`

Successfully connected to a Multidev without additional configuration 

## Additional information

### DIFF
--- a/starters/next-wordpress-starter/next.config.js
+++ b/starters/next-wordpress-starter/next.config.js
@@ -38,6 +38,20 @@ if (process.env.WPGRAPHQL_URL === undefined) {
 // remove trailing slash if it exists
 imageDomain = imageDomain.replace(/\/$/, '');
 
+let PANTHEON_ENVIRONMENT_PREFIX, IS_LIVE_ENVIRONMENT;
+if (process.env.PANTHEON_ENVIRONMENT_URL) {
+	const envPrefix = process.env.PANTHEON_ENVIRONMENT_URL.match(/^([^-]*)/)[0];
+	PANTHEON_ENVIRONMENT_PREFIX =
+		process.env.PANTHEON_ENVIRONMENT_URL.match(/^([^-]*-)[^-]*/)[0];
+	if (envPrefix !== 'live') {
+		PANTHEON_ENVIRONMENT_PREFIX =
+			process.env.PANTHEON_ENVIRONMENT_URL.match(/^([^-]*-)[^-]*/)[0];
+	} else {
+		PANTHEON_ENVIRONMENT_PREFIX = 'live';
+		IS_LIVE_ENVIRONMENT = 'live';
+	}
+}
+
 const injectedOptions = {};
 if (process.env.PANTHEON_UPLOAD_PATH) {
 	injectedOptions['basePath'] = process.env.PANTHEON_UPLOAD_PATH;
--- a/web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
+++ b/web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
@@ -67,8 +67,8 @@ must be used inside of `next.config.js`.
   beginning prefix (live, pr-number (pr-12), branch-name (multi-integration)) of
   `PANTHEON_ENVIRONMENT_URL`. This is to act as a variable to key off of and
   source data from a specific BE environment
-- `IS_LIVE` - Automatically set to true if `PANTHEON_ENVIRONMENT_URL` is live
-  and false otherwise
+- `IS_LIVE_ENVIRONMENT` - Automatically set to true if
+  `PANTHEON_ENVIRONMENT_URL` is live and false otherwise
 
 Either the `PANTHEON_CMS_ENDPOINT` or `WPGRAPHQL_URL` will need to be set.
 
--- a/web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
+++ b/web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
@@ -67,6 +67,8 @@ must be used inside of `next.config.js`.
   beginning prefix (live, pr-number (pr-12), branch-name (multi-integration)) of
   `PANTHEON_ENVIRONMENT_URL`. This is to act as a variable to key off of and
   source data from a specific BE environment
+- `IS_LIVE` - Automatically set to true if `PANTHEON_ENVIRONMENT_URL` is live
+  and false otherwise
 
 Either the `PANTHEON_CMS_ENDPOINT` or `WPGRAPHQL_URL` will need to be set.
 
--- a/web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
+++ b/web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
@@ -27,7 +27,8 @@ IMAGE_DOMAIN=my-image-cdn.site
 If your site is translated and you would like the hreflang metadata set
 correctly, you may set `FRONTEND_URL` to the URL of your frontend site.
 
-If the`FRONTEND_URL` is not set, it will default to the value of `PANTHEON_ENVIRONMENT_URL`
+If the`FRONTEND_URL` is not set, it will default to the value of
+`PANTHEON_ENVIRONMENT_URL`
 
 ```
 FRONTEND_URL=https://my-frontend-site.pantheon.site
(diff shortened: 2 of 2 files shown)


starters/next-wordpress-starter/next.config.js
```diff
@@ -38,6 +38,20 @@ if (process.env.WPGRAPHQL_URL === undefined) {
 // remove trailing slash if it exists
 imageDomain = imageDomain.replace(/\/$/, '');
 
+let PANTHEON_ENVIRONMENT_PREFIX, IS_LIVE_ENVIRONMENT;
```
CobyPear, line 41: Use `PANTHEON_ENVIRONMENT`
```suggestion
let PANTHEON_ENVIRONMENT, IS_LIVE_ENVIRONMENT;
```

mitchellmarkoff: Updated, also removed a duplicate line I had left in from tested. 👍

starters/next-wordpress-starter/next.config.js
```diff
@@ -38,6 +38,17 @@ if (process.env.WPGRAPHQL_URL === undefined) {
 // remove trailing slash if it exists
 imageDomain = imageDomain.replace(/\/$/, '');
 
+let PANTHEON_ENVIRONMENT, IS_LIVE_ENVIRONMENT;
```
CobyPear, line 41: I would expect these to either be added to `nextConfig.env` or `process.env` for use throughout the application as needed.

web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
```diff
@@ -56,3 +57,48 @@ CLIENT_SECRET
...
+```
+
+To use `PANTHEON_ENVIRONMENT` and connect to a Multidev backend, this line could
+be added under the above logic and after `PANTHEON_ENVIRONMENT_PREFIX` has been
+defined
+
+```
+backendUrl = `https://${PANTHEON_ENVIRONMENT}-${process.env.WPGRAPHQL_URL.replace(/^https?:\/\//,'',)}`
+
+```
```
CobyPear, line 104: Looks like this is in the next+drupal docs section.
This should go in its own entry in the next+wp section instead.

CobyPear (changes requested):

web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
CobyPear: I think this can be refined a bit more. The `PANTHEON_CMS_ENDPOINT` is only set on the infra, so we should show how to mock that for local dev.

starters/next-wordpress-starter/next.config.js
```diff
@@ -38,6 +38,17 @@ if (process.env.WPGRAPHQL_URL === undefined) {
...
+if (process.env.PANTHEON_ENVIRONMENT_URL) {
+	const envPrefix = process.env.PANTHEON_ENVIRONMENT_URL.match(/^([^-]*)/)[0];
+	PANTHEON_ENVIRONMENT =
+		process.env.PANTHEON_ENVIRONMENT_URL.match(/^([^-]*-)[^-]*/)[0];
```
CobyPear, line 45: Not sure this is matching an environment like `multi-demo-test`

web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
CobyPear: yup!

starters/next-wordpress-starter/next.config.js
mitchellmarkoff: That is correct it would not match that right now. How can we differentiate between the ending of a multidev name and the beginning of the site name? Since everything is divided by `-` I'm not sure how we can tell the difference dynamically.

CobyPear: Good question. Not sure there is a good way? Might be a limitation we need to document.
@backlineint any suggestions?
Also while I'm pinging you in, should the prefix contain the `multi-` prefix or should we lop that off? I'm thinking this isn't much good for dev and test if you need to name your FE site branch `multi-test` and then this doesn't work.

CobyPear: So the case where the backend has a different site name, would we expect the user to lop off their FE site name from `PANTHEON_ENVIRONMENT`?

web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
```diff
@@ -56,3 +57,48 @@ CLIENT_SECRET
...
+		);
+}
+```
+
+To use `PANTHEON_ENVIRONMENT` and connect to a Multidev backend, this line could
+be added under the above logic and after `PANTHEON_ENVIRONMENT_PREFIX` has been
+defined
+
+```
+backendUrl = `https://${PANTHEON_ENVIRONMENT}-${process.env.WPGRAPHQL_URL.replace(/^https?:\/\//,'',)}`
```
backlineint, line 102: Does PANTHEON_ENVIRONMENT need the process.env prefix?

web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
backlineint: @mitchellmarkoff could we add a simple example here that uses IS_LIVE_ENVIRONMENT here? Something as simple as:

if (!process.env.IS_LIVE_ENVIRONMENT) {
  // set backend URL for your test environment.
}

web/docs/Frontend Starters/Next.js/Next.js + Drupal/setting-environment-variables.md
mitchellmarkoff: That is correct and the update is staged in the current set of revisions 👍

starters/next-wordpress-starter/next.config.js
mitchellmarkoff: @backlineint @CobyPear Just tried to implement this idea to take everything before `.appa.pantheon.site` and am running into an issue with testing. It seems Pantheon has a character limit on multidev names so taking the entire prefix + site name and using that in naming the multidev environment is not viable.
.
Screen Shot 2022-12-06 at 2 18 58 PM

Any thoughts on how to proceed?

CobyPear: Hmm yeah 11 chars is pretty short. This is making me wonder if we can ask the BE to expose this env var for us given the complexity of parsing it out. I don't remember if there was something preventing that on their side or if we just need to ask again.

mitchellmarkoff: I agree, assuming there is nothing preventing that on their side I think that is the smoothest path forward. Should I move this ticket to blocked and add a note?

CobyPear: That makes sense to me

starters/next-wordpress-starter/next.config.js
```diff
@@ -38,6 +38,13 @@ if (process.env.WPGRAPHQL_URL === undefined) {
 // remove trailing slash if it exists
 imageDomain = imageDomain.replace(/\/$/, '');
 
+if (process.env.PANTHEON_ENVIRONMENT_URL) {
+	const envPrefix = process.env.PANTHEON_ENVIRONMENT_URL.match(/^([^-]*)/)[0];
+	if (envPrefix === 'live') {
+		process.env.IS_LIVE_ENVIRONMENT = 'live';
+	}
```
CobyPear, line 45: If we're just checking for `live` we can replace the regex with `startsWith()`

web/docs/Frontend Starters/Next.js/Next.js + WordPress/setting-environment-variables.md
```diff
@@ -23,3 +23,61 @@ For example:
...
+		process.env.WPGRAPHQL_URL.replace(/\/wp\/graphql$/, '').replace(
+			/^https?:\/\//,
+			'',
+		);
+}
+```
+
+In order to connect to a Multidev backend, this backendUrl will need to be
+updated. `PANTHEON_ENVIRONMENT_URL` has the data needed to source your backend
+and will need to be parsed according to your needs.
```
CobyPear, line 68:
```suggestion
updated. `PANTHEON_ENVIRONMENT_URL` includes the PR number or integration branch name
and will need to be parsed according to your needs.
```
Doesn't have to be the suggestion, but I think we need something less vague than 'has the data needed'

web/docs/Frontend Starters/Next.js/Next.js + WordPress/setting-environment-variables.md
```diff
@@ -23,6 +23,64 @@ WPGRAPHQL_URL=https://my-wordpress-site.pantheon.site/wp/graphql
 IMAGE_DOMAIN=my-image-cdn.site
 ```
 
+## Connecting to Multidev Cloud Environments
```
CobyPear, line 26:
```suggestion
## Connecting to Multidev Environments
```

starters/next-wordpress-starter/next.config.js
```diff
@@ -38,6 +38,12 @@ if (process.env.WPGRAPHQL_URL === undefined) {
 // remove trailing slash if it exists
 imageDomain = imageDomain.replace(/\/$/, '');
 
+if (process.env.PANTHEON_ENVIRONMENT_URL) {
+	if (process.env.PANTHEON_ENVIRONMENT_URL.startsWith('live') === true) {
+		process.env.IS_LIVE_ENVIRONMENT = 'live';
```
CobyPear, line 43: Technically any string will work here, but let's set it to 'true' for consistency with the docs

web/docs/Frontend Starters/Next.js/Next.js + WordPress/setting-environment-variables.md
```diff
@@ -23,6 +23,64 @@ WPGRAPHQL_URL=https://my-wordpress-site.pantheon.site/wp/graphql
...
+
+To connect to a Multidev environment, the following helper environment variable
+can be used inside of `next.config.js`.
+
+- `IS_LIVE_ENVIRONMENT` - Automatically set to true if
+  `PANTHEON_ENVIRONMENT_URL` is live and false otherwise
```
CobyPear, line 32:
```suggestion
- `IS_LIVE_ENVIRONMENT` - True if
  `PANTHEON_ENVIRONMENT_URL` is live.
```

web/docs/Frontend Starters/Next.js/Next.js + WordPress/setting-environment-variables.md
```diff
@@ -23,6 +23,64 @@ WPGRAPHQL_URL=https://my-wordpress-site.pantheon.site/wp/graphql
...
+
+Either the `PANTHEON_CMS_ENDPOINT` or `WPGRAPHQL_URL` will need to be set.
+
+The `PANTHEON_CMS_ENDPOINT` can be mocked for local development by defining it
+in the `.env.development.local`
+
+```
+PANTHEON_CMS_ENDPOINT=
```
CobyPear, line 40: Add an example here. This is needed because the `PANTHEON_CMS_ENDPOINT` is not the same format as the `WPGRAPHQL_URL`
```suggestion
PANTHEON_CMS_ENDPOINT=dev-my-wordpress-site.pantheonsite.io
```

starters/next-wordpress-starter/.env.example
```diff
@@ -1,6 +1,8 @@
 WPGRAPHQL_URL=
 IMAGE_DOMAIN=
 #PANTHEON_UPLOAD_PATH=/example
-#PREVIEW_SECRET=
+PREVIEW_SECRET=
```
CobyPear, line 4: I like having the optional vars commented out. Maybe we can even add a note at the top explaining that?
Or we should uncomment all of them.
My vote is to comment out this and `IS_LIVE_ENVIRONMENT`

CobyPear (changes requested):

starters/next-wordpress-starter/.env.example
```diff
@@ -1,6 +1,11 @@
+/**
+* Optional environment variables are prefaced with `#`
+**/
```
CobyPear, line 3: .env files use `#` for comments
```suggestion
# Optional environment variables are commented out
```

CobyPear (approved):
:shipit:

PR #452, opened 2022-12-13
DB-4935: Error if previewing a draft

CobyPear:
## What changes were made?
Drafts are not currently supported for Drupal's decoupled_preview. This change adds a clearer error message in the case a user tries to preview a draft.

next-drupal-starter

locally
## Additional information

### DIFF
--- a/starters/next-drupal-starter/lib/getPreview.js
+++ b/starters/next-drupal-starter/lib/getPreview.js
@@ -34,7 +34,7 @@ export async function getPreview(context, node, params) {
 			if (context?.previewData?.resourceVersionId) {
 				process.env.DEBUG_MODE &&
 					console.log(
-						`Adding resource version ID param ${context?.previewData?.resourceVersionId}...`,
+						`Adding resource version ID param ${context.previewData.resourceVersionId}...`,
 					);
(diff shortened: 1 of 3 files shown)


starters/next-drupal-starter/lib/getPreview.js
```diff
@@ -51,10 +51,18 @@ export async function getPreview(context, node, params) {
 					}`,
 					requestInit,
 				);
+				if (!fetchedPreviewData?.data?.id) {
+					throw new Error(
+						'Unable to fetch preview data. Drafts that have been saved are not preview-able and must be published again before preview is possible.',
```
backlineint, line 56: This message will be awkward regardless, but what about:

Unable to fetch preview data. Previewing drafts is currently not supported. As an alternative, either preview while editing, or preview a saved revision.

backlineint (changes requested):

starters/next-drupal-starter/lib/getPreview.js
CobyPear: Works for me! change inc.

backlineint (approved):

PR #471, opened 2023-01-17
DB-5231: Support GraphQL over HTTP GET

mitchellmarkoff:
## What changes were made?
- Support GraphQL over HTTP GET added to the `wordpress-kit`

- `wordpress-kit`

- A GraphQL client was created to support GET requests and the use of this client resulted in successful cache hits.
-   
## Additional information

### DIFF
--- a/packages/wordpress-kit/__tests__/GraphqlClientFactory.test.ts
+++ b/packages/wordpress-kit/__tests__/GraphqlClientFactory.test.ts
@@ -30,7 +30,6 @@ test('request and response middleware add appropriate headers', async () => {
 });
 
 test('client is configured for GET requests', async () => {
-	// const client = new GraphqlClientFactory('http://my-wp.test/graphql').create();
 	const getClient = new GraphqlClientFactory('http://my-wp.test/graphql', {
 		method: 'GET',
 	}).create();
--- a/packages/wordpress-kit/__tests__/GraphqlClientFactory.test.ts
+++ b/packages/wordpress-kit/__tests__/GraphqlClientFactory.test.ts
@@ -28,3 +28,12 @@ test('request and response middleware add appropriate headers', async () => {
 	expect(res.data).toEqual(basicPostsQuery);
 	expect(res.headers.has('surrogate-key-raw')).toBeTruthy();
 });
+
+test('client is configured for GET requests', async () => {
+	const client = new GraphqlClientFactory('http://my-wp.test/graphql').create();
+	const getClient = new GraphqlClientFactory('http://my-wp.test/graphql', {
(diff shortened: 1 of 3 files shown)


mitchellmarkoff:
@CobyPear LMK your thoughts in relation to this work and documentation. I didn't update docs here since `GET` is passed through the options param which is [already documented](https://live-decoupled-kit-docs-canary.appa.pantheon.site/docs/Packages/wordpress-kit/classes/GraphqlClientFactory). Also figured implementation examples would go into the frontend starter docs as opposed to the package docs.

CobyPear:
> Also figured implementation examples would go into the frontend starter docs as opposed to the package docs.

That sounds accurate. Any reason not to do that with this ticket?

mitchellmarkoff:
> That sounds accurate. Any reason not to do that with this ticket?

 I held off on it because of this point in the implementation details thinking that docs updates would come during the starter related work. 
`Changes here are only for wordpress-kit. This will be consumed in the starter kits as part of another ticket.`
With that said, the docs should be simple to write  so if it's easier to add them with this PR that is something I can do.

CobyPear:
> > That sounds accurate. Any reason not to do that with this ticket?
> 
> I held off on it because of this point in the implementation details thinking that docs updates would come during the starter related work. `Changes here are only for wordpress-kit. This will be consumed in the starter kits as part of another ticket.` With that said, the docs should be simple to write so if it's easier to add them with this PR that is something I can do.

oh good call, missed that bit. You can save it for the other ticket. This one should have a test at the very least tho.

packages/wordpress-kit/__tests__/GraphqlClientFactory.test.ts
```diff
@@ -28,3 +28,12 @@ test('request and response middleware add appropriate headers', async () => {
...
 	expect(res.headers.has('surrogate-key-raw')).toBeTruthy();
 });
+
+test('client is configured for GET requests', async () => {
+	const client = new GraphqlClientFactory('http://my-wp.test/graphql').create();
+	const getClient = new GraphqlClientFactory('http://my-wp.test/graphql', {
+		method: 'GET',
+	}).create();
+
+	expect(client).not.toMatchObject(getClient);
```
CobyPear, line 38: We can be a bit more specific here:
```suggestion
	const getClient = new GraphqlClientFactory('http://my-wp.test/graphql', {
		method: 'GET',
	});

	expect(getClient.options.method).toEqual('GET');
```

CobyPear (changes requested):

CobyPear (approved):

PR #503, opened 2023-02-01
DB-5052: Gatsby Wordpress Generator

mitchellmarkoff:
## What changes were made?
Added the `gatsby-wordpress-starter` as a generator to the cli.

create-pantheon-decoupled-kit

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/package.json.hbs
```diff
@@ -0,0 +1,61 @@
...
+		"test:watch": "vitest",
+		"update-snapshots": "vitest run --update --silent",
+		"coverage": "vitest run --coverage"
+	},
+	"dependencies": {
+		"gatsby": "^4.24.0",
+		"gatsby-plugin-image": "^2.24.0",
+		"gatsby-plugin-manifest": "^4.24.0",
+		"gatsby-plugin-offline": "^5.24.0",
+		"gatsby-plugin-pnpm": "^1.2.10",
```
CobyPear, line 30: It would be great if we could only add this dependency if the package manager is `pnpm`. We already have the `which-pm-runs` dependency added. You should be able to use that dependency in the gatsby-wp generator to determine if this package should be added or not.

CobyPear (changes requested):
One other thing to add is a `.env.development.local.hbs`. I haven't checked yet, but we may need to refactor the `runESLint` action for this one, as I don't think the gatsby starter comes with a linit script like nextjs does.

packages/create-pantheon-decoupled-kit/src/generators/gatsby-wp.generator.ts
```diff
@@ -0,0 +1,47 @@
...
+			default: ({ appName }: { [key: string]: string }) =>
+				`${process.cwd()}/${appName.replaceAll(' ', '-').toLowerCase()}`,
+		},
+	],
+	actions: (data) => {
+		let pnpm = false;
+		const getPkgManager = whichPmRuns();
+		if (getPkgManager && getPkgManager.name === 'pnpm') {
+			pnpm = true;
+		}
```
CobyPear, line 26: This could be a bit more succinct
```suggestion
		const pnpm = whichPmRuns()?.name === 'pnpm' ? true : false;

```

packages/create-pantheon-decoupled-kit/src/utils/runESLint.ts
```diff
@@ -25,9 +25,14 @@ export const runESLint = (
...
-		if (error instanceof Error) {
-			throw error;
+		try {
+			execSync('npx eslint', { cwd: answers.outDir, stdio: 'inherit' });
+			return 'success eslint';
+		} catch (error) {
+			if (error instanceof Error) {
+				throw error;
+			}
+			throw 'fail';
```
CobyPear, line 35: Not a huge fan of this pattern as it will still show some nasty errors in the console. My suggestion would be to read from the `scripts` of the new`package.json` to see if there is a `lint` script, then call that script if it exists, otherwise use `npx eslint --write ` with configurable options like ignore pattern/path or plugins. Meaning, I should be able to pass options ito the `runESLint` config that are then passed into this command.

CobyPear (changes requested):
The pnpm plugin conditional is working well. I'm not so sure on the nested try/catch though. See comments.

packages/create-pantheon-decoupled-kit/src/utils/runESLint.ts
```diff
@@ -21,12 +31,24 @@ export const runESLint = (
...
 		command = getPkgManager.name;
 	}
-	try {
-		execSync(`${command} lint`, { cwd: answers.outDir, stdio: 'inherit' });
-	} catch (error) {
-		if (error instanceof Error) {
-			throw error;
+
+	const filePath = path.resolve(`${answers.outDir}/package.json`);
+	const script = JSON.parse(fs.readFileSync(filePath, 'utf8')) as Script;
```
CobyPear, line 36: Can we be a bit more specific with the names here?
```suggestion
	const pkgPath = path.resolve(`${answers.outDir}/package.json`);
	const pkg = JSON.parse(fs.readFileSync(filePath, 'utf8'));
```

I'm wondering if it would be better to break this functionality out since we may want to reuse this pattern in the future. See https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/common/utils/helpers.ts#L33 for an example of how this could look. That way we could have a more robust type for the `package.json`.

packages/create-pantheon-decoupled-kit/src/utils/runESLint.ts
```diff
@@ -21,12 +31,24 @@ export const runESLint = (
...
+	if (script.scripts.lint) {
+		try {
+			execSync(`${command} lint`, { cwd: answers.outDir, stdio: 'inherit' });
```
CobyPear, line 39: Unsure this does anything without providing file extensions or paths to lint. Need to do some more validation here but I think we at least need `npm run lint .` to tell the cli where/what to lint.

packages/create-pantheon-decoupled-kit/__tests__/runESLint.test.ts
```diff
@@ -12,15 +14,48 @@ vi.mock('which-pm-runs');
...
+		['package.json', 'node_modules'].forEach((path) => {
+			fs.rmSync(`${outDir('withLint')}/${path}`, {
+				force: true,
+				recursive: true,
+			});
+			fs.rmSync(`${outDir('withoutLint')}/${path}`, {
+				force: true,
+				recursive: true,
+			});
+		});
```
CobyPear, line 54: Not sure we need this pattern here since the tests aren't manipulating the files in any way like the other tests are. I think you could just commit the fixtures instead of programmatically creating the files before/after.

CobyPear (changes requested):
I think we could maybe break out the package.json read stuff into it's own utility so that we could reuse it and get a more robust type.

One other interesting thing that I found is that we need to change the root `.prettierignore`.
This line:
```
**/templates/**
```
makes it so that a generated gatsby starter inside the monorepo will skip linting for the `templates` directory in the gatsby starter. 

I think we can be less specific on the glob and it will fix it
```
**/create-pantheon-decoupled-kit/**/templates/**
```

packages/create-pantheon-decoupled-kit/__tests__/runESLint.test.ts
mitchellmarkoff: @CobyPear I ended up going with this pattern because committing `package.json` files to the fixtures proved to be troublesome. Running `pnpm test` with the committed files fails and requires `node_modules` to be installed at the file path. After installing `node_modules`,  `pnpm test` continues to fail since it then treats those folders as their own projects which has missing test coverage. Below is the error output. LMK if there is a better solution but I found programmatically creating the files before/after to be an easy workaround.

Screen Shot 2023-02-07 at 2 25 36 PM

CobyPear: I did not have the same error, I was able to run tests with a `package.json` inside the fixtures.
I do see them being picked up by the monorepo workspace, but we can add another filter in the `pnpm-workspace.yml` to prevent this:
`  - '!packages/create-pantheon-decoupled-kit/__tests__/**'`

CobyPear: looking a bit closer at your screenshot, looks like you may be in a subdir and `vitest` isn't finding any tests.

packages/create-pantheon-decoupled-kit/src/utils/runESLint.ts
mitchellmarkoff: When breaking this functionality out, is it alright to use the `any` type similar to the Sitecore example?

mitchellmarkoff:
LGTM!

CobyPear (approved):
:shipit:

PR #510, opened 2023-02-07
DB-940: Document Build Hooks Configuration.

abhisekmazumdar:
## What changes were made?
Created a new page for documentation of Build Hooks Configuration.

Documentation


web/docs/Backend Starters/Decoupled Drupal/configuring-build-hooks.md
```diff
@@ -0,0 +1,49 @@
...
+in a decoupled frontend site.
+
+### Installing the Build Hooks Module
+
+To add Build Hooks as a composer dependency, run the following command:
+
+```
+composer require 'drupal/build_hooks:^3.3'
+```
+
```
backlineint, line 23: It might also be worth adding a note that by default the module will only log changes for content. If you'd like to trigger build hooks based on other entity updates,  Administration -> Configuration ->  Build hooks -> Build Hooks Settings.

web/docs/Backend Starters/Decoupled Drupal/configuring-build-hooks.md
```diff
@@ -0,0 +1,49 @@
...
+### Creating a Build Hook for Content Creation/Modification
```
backlineint, line 24: It is more verbose, but I think it will be more useful to split this out into two full sections:

Creating a Build Hook Triggered Automatically When Content Changes

Creating a Build Hook Triggered Manually

Each will include the steps for picking a deployment strategy and testing (probably not as their own sections) Many of the instructions will be the same, but for the manual one, we can go into more detail about where the user will need to go in the Drupal admin, and how they can trigger the hook manually.

I don't think we need a full Cron example, but maybe we could note that deployment strategy is also an option.

backlineint (changes requested):
I'd like to see us go into a little more detail here about the two main use cases. I think it will make a big difference in helping people understand what is possible with build hooks. Ideally we'd be referencing the build hooks docs more here, but as you know, they aren't that detailed.

web/docs/Backend Starters/Decoupled Drupal/configuring-build-hooks.md
```diff
@@ -0,0 +1,82 @@
...
+
+This will automatically trigger a build hook when content changes.
+
+### Trigger a Deployment Manually
+
+Following the same steps as "Trigger a Deployment Automatically" to create a
+Build Hook Frontend Environment but select **Deployment strategy** as **Manually
+only**.
+
+![Drupal Build Hook Trigger Deployment Manually](../../../static/img/drupal-trigger-deployment-manually.gif)
```
abhisekmazumdar, line 70: I feel the images location set here might break the final build. I tried finding any example to do it correctly. But had no luck.
Kindly let me know if this needs to be change in any way?

backlineint: Seems to be working for me locally...

backlineint (approved):

PR #512, opened 2023-02-07
DB-5376: Update Default Cache Control Values

mitchellmarkoff:
## What changes were made?
- Server side rendered responses now use a Cache-Control header of public, s-maxage=600
- Documentation updated accordingly.

`drupal-kit`
`wordpress-kit`
./web

Starter observed to run normally with updated headers.

## Additional information

### DIFF
--- a/.changeset/hip-seas-beam.md
+++ b/.changeset/hip-seas-beam.md
@@ -1,5 +1,7 @@
 ---
-"@pantheon-systems/wordpress-kit": patch
+'@pantheon-systems/wordpress-kit': minor
+'@pantheon-systems/drupal-kit': minor
 ---
 
-Update Default Cache Control Values
+Updated default Cache-Control value from
+`public, s-maxage=10, stale-while-revalidate=600` to `s-public, maxage=600`
--- a/.changeset/hip-seas-beam.md
+++ b/.changeset/hip-seas-beam.md
@@ -1,0 +1,5 @@
+---
+"@pantheon-systems/wordpress-kit": patch
+---
+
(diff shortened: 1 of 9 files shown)


mitchellmarkoff:
Made this a `minor` version bump since functionality is changing, could see a case for `patch` as well.

.changeset/hip-seas-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pantheon-systems/wordpress-kit": patch
```
CobyPear, line 2: I see this one is a patch and the other is a minor version. I think minor version makes sense. For this, since it's the same functionality across both packages, you can use a single changeset file.

.changeset/hip-seas-beam.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Update Default Cache Control Values
```
CobyPear, line 5: Add the old and new values
```suggestion
Updated default Cache-Control value from `public, s-maxage=10, stale-while-revalidate=600` to `s-public, maxage=600`
```

CobyPear (changes requested):
code and docs LGTM, just an update to the changesets.

CobyPear:
> Made this a `minor` version bump since functionality is changing, could see a case for `patch` as well.

Hmm I said to use minor in my other comment but thinking about it a bit more I think patch is correct here. This could be considered a bug fix or optimization which is backwards compatible.

CobyPear (approved):

PR #515, opened 2023-02-10
DB-5253: Part 1 - initial cleanup

CobyPear:
## What changes were made?

- Remove root CHANGELOG.md
  - This hasn't really been in use anyway and now seems like an OK time to clean it up
- Add pkgName hbs helper to properly handle package.json names
- Add partial for shared package.json fields
- Move actions to src/actions directory
- Skip lint if noInstall or noLint is in answers
- Enhance some TSDoc comments and regenerate cli api ref
- Add unit test for pkgNameHelper
- Add help and version commands
- Add changeset

Somewhat related to this work, I noticed some issues with the lint command when there is no lint script, but I think we should spend some time discussing the best approach for post-project generation linting/formatting. `npx eslint` doesn't work without a config present. This is not a huge concern or blocker at the moment, but we should get it worked out fairly soon.

create-pantheon-decoupled-kit

Tested all new helpers and commands locally

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/package.json.hbs
```diff
@@ -1,13 +1,5 @@
...
-	"name": "{{kebabCase appName}}",
-	"version": "0.1.0",
-	"private": true,
-	"license": "GPL-3.0-or-later",
-	"description": "",
-	"homepage": "",
-	"bugs": "",
-	"author": "",
-	"repository": "",
+	{{> sharedPkgJsonFields}}
```
backlineint, line 2: Only noticed one tiny thing when reviewing this. The introduction of the partial here is resulting in the scripts section not starting on a new line. Still valid JSON, but something we should clean up.

Screenshot at Feb 13 10-32-20

backlineint (changes requested):

packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/package.json.hbs
CobyPear: nice catch! this should be fixed now.

backlineint (approved):

PR #520, opened 2023-02-14
DB-5053: Umami Generator Template

mitchellmarkoff:
## What changes were made?
Umami data and routes added as a umami-demo template.

Both generators successfully used alongside each other to create a project with the umami routes.

## Additional information


packages/create-pantheon-decoupled-kit/src/generators/umami.generator.ts
```diff
@@ -0,0 +1,28 @@
...
+import type { DecoupledKitGenerator } from '../types';
+
+export const umami: DecoupledKitGenerator = {
+	name: 'umami-demo',
+	description: 'Umami profile data and components',
+	prompts: [
+		{
+			name: 'outDir',
+			message: 'Where should the output go?',
+		},
```
CobyPear, line 11: this is missing the default output. Check the other generators outDir prompt, it should be the same here.

packages/create-pantheon-decoupled-kit/src/generators/umami.generator.ts
```diff
@@ -0,0 +1,28 @@
+import type { CustomActionConfig } from 'node-plop';
+import type { DecoupledKitGenerator } from '../types';
+
+export const umami: DecoupledKitGenerator = {
+	name: 'umami-demo',
+	description: 'Umami profile data and components',
```
CobyPear, line 6:
```suggestion
	description: 'Drupal's Umami profile data and components add-on for the next-drupal starter',
```
The description could be a bit more precise. I am now realizing that there is no place to see these descriptions... Maybe they should show up with the list of generators? You don't need to implement this in this ticket, just thinking out loud. Thoughts?

CobyPear (changes requested):
Left some inline comments.

When I tried to select both generators at one time, it highlighted a problem that in hindsight should have been obvious - we need to pool all of the actions together before calling them, so that the first generator doesn't bother trying to install and lint until all templates have been rendered and written to their appropriate place.

I will think on this a bit more. Not sure if this logic should be wrapped around plop somehow, or if in this case we need to go around plop like we are with the prompts.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-umami/pages/[...alias].jsx
```diff
@@ -0,0 +1,189 @@
...
+					instructions={field_recipe_instruction.value}
+				/>
+			);
+		}
+		return (
+			<>
+				<h2 className="text-xl text-center mt-14">No content found 🏜</h2>
+			</>
+		);
+	};
```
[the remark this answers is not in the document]
Comment (mitchellmarkoff) [reply packages/create-pantheon-decoupled-kit/src/templates/next-drupal-umami/pages/[...alias].jsx:97]: @CobyPear I see how this could reduce the number of individual template files, however I do have some implementation concerns with this.

If some umami data is wrapped in a handlebars conditional inside of the next-drupal template, wouldn't that defeat the purpose of this addon? With that approach, in order to add umami data into an  preexisting project, a user would need to run both generators to get all of the umami data. Rerunning both of the generators would also overwrite any changes someone would have made to their project.

I can see this working if we are only giving the option to add umami data during the initial project generation but otherwise it makes the addon harder and more confusing to use IMO.

Thoughts?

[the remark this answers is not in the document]
Comment (CobyPear) [reply packages/create-pantheon-decoupled-kit/src/templates/next-drupal-umami/pages/[...alias].jsx:97]: > With that approach, in order to add umami data into an preexisting project, a user would need to run both generators to get all of the umami data.

I don't see why that would be true. For example, if you configure the generator to pass in `umami: true` to the `data` in the actions, the templates can render only the pieces in the `umami` conditional.

>  overwrite any changes someone would have made to their project.

This is why we have the diff and prompt on each change.

packages/create-pantheon-decoupled-kit/src/generators/umami.generator.ts
mitchellmarkoff: I set a new default for this `outDir` that is based on the other generators `outDir` prompts. It is slightly different since this generator does not take an `appName`

.changeset/three-meals-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-pantheon-decoupled-kit": minor
+---
+
+Umami Generator Template
```
CobyPear, line 5:
```suggestion
Added Drupal Umami demo examples as a generator
```

CobyPear (changes requested):
LGTM, two small things:
1. changeset could be a bit more descriptive
2. There is one change to `src/index.ts` that I think should be made here taken from that snipped I sent earlier. specifically the part which adds the `answers` from the prompts back to the `args` object. Then use `args` instead of `answers` when calling `runActions`

CobyPear (approved):
I changed the name of the generator to something a bit more descriptive. Everything is working fine, GJ!

PR #537, opened 2023-02-21
DB-5403: `next-wp` Related Content ACF Component Addon

mitchellmarkoff:
## What changes were made?
- Related Content component and query addon created in the CLI

Both generators successfully used alongside each other to create a project with ACF support.

## Additional information


packages/create-pantheon-decoupled-kit/src/generators/next-wp-acf-addon.generator.ts
```diff
@@ -0,0 +1,30 @@
+import type { CustomActionConfig } from 'node-plop';
+import type { DecoupledKitGenerator } from '../types';
+
+export const nextWpAcfAddon: DecoupledKitGenerator = {
+	name: 'next-wp-acf-addon',
+	description:
+		"WordPress's advance custom field component add-on for the next-wordpress starter",
```
CobyPear, line 7:
```suggestion
		"Example implementation of the WordPress Advanced Custom Fields plugin for the next-wordpress starter",
```

CobyPear (changes requested):
Looking good. Just a few small things.
Also, there shouldn't be too much to refactor when we change to the new action runner which will be nice.

.changeset/wild-mails-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-pantheon-decoupled-kit": minor
+---
+
+Related content ACF component addon for next-wp
```
CobyPear, line 5: Sorry, missed this the first time:
Think about how this will look on the changelog when we do a release. In this case I think it could be formatted a bit differently so that it tells a reader not familiar with ACF or an addon what this feature does.
```suggestion
Added a new addon, `next-wp-acf-addon`. It includes an example implementation of data sourced from WordPress with the [Advanced Custom Fields plugin](https://www.advancedcustomfields.com/) activated.
```

CobyPear (changes requested):
One comment re: changeset that I missed the first time. Code changes look good to me!

CobyPear (approved):

PR #545, opened 2023-02-24
DB-5455: Advanced Custom Fields Add-on Docs

mitchellmarkoff:
## What changes were made?
- A new entry titled ‘Add-ons’ is added to the Decoupled Kit developer docs under Front-End Starters → Next.js → Next.js and WordPress and also under Front-End Starters → Gatsby and WordPress.

`./web`

Ran docs site locally.

## Additional information


web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
+---
+id: 'gatsby-wordpress-add-ons'
+title: 'Add Ons'
```
CobyPear, line 3:
```suggestion
title: 'Add-ons'
```
'Add-ons' should be hyphenated in this case.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+slug: '/frontend-starters/nextjs/gatsby/gatsby-wordpress/add-ons'
+sidebar_position: 5
+---
+
+### Advanced Custom Fields
```
CobyPear, line 8:
```suggestion
## Advanced Custom Fields WordPress Plugin
```
use an h2 here, also probably doesn't hurt to use the full name

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+posts are specified on the related WordPress post.
+
+#### Before You Begin
+
+- Install and activate the
+  [WP Advanced Custom Fields Plugin](https://wordpress.org/plugins/advanced-custom-fields/).
+
+- Install the
+  [Create Pantheon Decoupled Kit CLI](https://www.npmjs.com/package/create-pantheon-decoupled-kit/).
+
```
CobyPear, line 29: "Before You Begin" should go at the top, also as an h2.

there should be no need to actually install the decoupled-kit cli. `npm create pantheon-decoupled-kit` (or pnpm create) pull the latest binary from the registry and execute it. We should probably stick with `npm create` for these docs, or mention that it will work with 'your preferred package manager (npm, pnpm, yarn)'

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
+---
+id: 'gatsby-wordpress-add-ons'
+title: 'Add Ons'
+slug: '/frontend-starters/nextjs/gatsby/gatsby-wordpress/add-ons'
+sidebar_position: 5
+---
+
+### Advanced Custom Fields
+
+Advanced Custom Fields turns WordPress sites into a fully-fledged content
```
CobyPear, line 10:
```suggestion
The Advanced Custom Fields WordPress plugin, or ACF, turns WordPress sites into a fully-fledged content
```
introduce the full name the first time with the abbreviation, so from here on you can just use 'ACF' instead of spelling out Advanced Custom Fields each time.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+management system by giving you all the tools to do more with your data.
+
+Advanced custom fields let you add, store, and display additional information
+about a piece of content in WordPress. This information can easily be assessed
+through GraphQL with simple adjustments to existing queries.
+
+This add-on uses GraphQL to bring in a custom `related content` data type that
+was created with Advanced Custom Fields. Your starter kit will have a related
```
CobyPear, line 18:
```suggestion
was created with Advanced Custom Fields. After adding the `gatsby-wp-acf-addon`, our starter kit will have a related
```
More specificity here - otherwise it reads as though the starter kit has this already.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Before You Begin
+
+- Install and activate the
+  [WP Advanced Custom Fields Plugin](https://wordpress.org/plugins/advanced-custom-fields/).
+
+- Install the
+  [Create Pantheon Decoupled Kit CLI](https://www.npmjs.com/package/create-pantheon-decoupled-kit/).
+
+#### Instillation Through CLI
```
CobyPear, line 30:
```suggestion
## Adding the `gatsby-wp-acf-addon`  to an Existing Project
```
'Installing' doesn't feel right to me, open to other options here. Also, this should either be a new section (h2), or be nested under a more generic heading like "Setup" or move "Usage" up to here as the generic section

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+
+1. Use the create command to initiate the cli
+   `pnpm create pantheon-decoupled-kit`
```
CobyPear, line 33: As mentioned, use `npm` here. Maybe in Before You Begin, mention that any (npm, pnpm, yarn) package manager will work.
You could include the add-on as an arg to skip step 2
`npm create pantheon-decoupled-kit gatsby-wp-acf-addon`

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+1. Select the `gatsby-wp-acf-addon` and go through the prompts. Set the output
+   to the directory that your Gatsby Wordpress Starter is located
+1. Once the app is generated, add
```
CobyPear, line 36: This line reads as though we are generating an app from 0, not adding an add-on. I think it's fine to describe both use cases, but it should be clear what the starting point is.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+   `https://acf-decoupled-wordpress-qa.pantheonsite.io/wp/graphql` as the
```
CobyPear, line 37: Not sure we want to publicize this URL 😬

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,96 @@
...
+
+##### Adding Custom Field Data
+
+Once you have created a field group, values can be assigned to your new custom
+fields via the Guttenberg editor.
+
+1. Select and existing post or page, or create a new one.
+1. At the bottom of your editor, you will see the new custom fields that have
+   been created, fill them in with data of your choosing
+1. Save and publish your content updates
```
CobyPear, line 64: Not sure this makes sense in this particular doc. I would expect that a user has some understanding of ACF from the WP side if they're using this add-on. This doc should concern itself more with how to fetch the data, not how to create it.

CobyPear (changes requested):
All of the comments on the Gatsby version of the doc should apply to the Next.js version too.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,87 @@
...
+
+- Familiarize yourself with the
+  [Create Pantheon Decoupled Kit CLI](https://www.npmjs.com/package/create-pantheon-decoupled-kit/).
+  Any package manager can be used with the CLI **[npm, pnpm, yarn]**.
+
+## Advanced Custom Fields WordPress Plugin
+
+The Advanced Custom Fields WordPress plugin, or ACF, turns WordPress sites into
+a fully-fledged content management system by giving you all the tools to do more
+with your data.
```
CobyPear, line 21: Two things:
1. If you take text directly from a source on the web, add an attribution to that source.
2. I want to avoid explaining what ACF is in our docs. I would expect someone reading this to already be familiar with ACF, and if not we can link to their docs so the user can learn about it before coming back here and integrating ACF with our starter via the add-on. So in that case, I would prefer a note in "Before You Begin" that explains "You should be familiar with WordPress Advanced Custom Fields" with a link to the docs. I don't think we should explain what it is here.

This section should explain what our add-on does, but we should avoid explaining ACF.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Adding the `gatsby-wp-acf-addon` to an Existing Project
+
+1. Use the create command to initiate the cli with the `gatsby-wp-acf-addon`
+   generator
+   ```bash
+   npm create pantheon-decoupled-kit gatsby-wp-acf-addon
+   ```
+1. When prompted for an output directory, set the output to the root directory
+   of your existing Gatsby Wordpress Starter
```
CobyPear, line 42:
```suggestion
   of your existing Gatsby WordPress Starter
```

CobyPear (changes requested):
I'm still of the opinion that we don't need to explain ACF and we expect a user to already be familiar with it if they're using this add-on. Also, please add attribution and/or a link to the original content when using text from other sites.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/add-ons.md
CobyPear: @mitchellmarkoff Going back to this for a second – I am realizing this doc needs to be structured slightly differently because it will relate to multiple add-ons. I'm not sure this will be the structure we use going forward but for now I guess it makes sense.
Because of that, the `h2`s here should be the addon name and under it will be the description. Then `h3`s under that should start each section like Before You Begin, Usage, and so on. I've pushed up this change, let me know if it makes sense to you.

mitchellmarkoff: Makes total sense. I was actually wondering how formatting would work when the time would come to add to this doc, I should have started that conversation earlier in the process. LGTM!

CobyPear (approved):

PR #546, opened 2023-02-24
DB-5494: Publish Prettier and Eslint Configs

mitchellmarkoff:
## What changes were made?
The decoupled-kit-js monorepo configs are now in their own package to get published to npm

`configs`

The configs have been observed to continue to work inside of the monorepo

## Additional information


mitchellmarkoff:
I ended up needing to publish the two configurations separately in the end. Let me know if this changes any ideas with naming conventions or setup.

.changeset/curly-insects-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pantheon-systems/configs": minor
+---
+
+Publish monorepo prettier configs
```
CobyPear, line 5: The local `configs` workspace should not be published or need a changeset.

mitchellmarkoff:
@CobyPear Putting the local .`/configs` workspace back in led to some routing issues since there were then two `@pantheon-systems/configs` and `@pantheon-systems/eslint-config` packages present in the mono repo. To get around this I renamed the two new packages to `@pantheon-systems/style-configs` and `@pantheon-systems/style-eslint-config`. LMK if you have any thoughts on this.

mitchellmarkoff:
This is my first time publishing a new package and I was not sure if the release of a package is automatically considered a major release or not. I ended up going with a minor version here. LMK if you have any input on versioning.

CobyPear:
> This is my first time publishing a new package and I was not sure if the release of a package is automatically considered a major release or not. I ended up going with a minor version here. LMK if you have any input on versioning.

My understanding of semver is that we should start at `0.1.0` and work towards a more stable `1.0.0` version.
https://semver.org/#spec-item-4

CobyPear:
> @CobyPear Putting the local .`/configs` workspace back in led to some routing issues since there were then two `@pantheon-systems/configs` and `@pantheon-systems/eslint-config` packages present in the mono repo. To get around this I renamed the two new packages to `@pantheon-systems/style-configs` and `@pantheon-systems/style-eslint-config`. LMK if you have any thoughts on this.

 It's possible there is a better structure for this to begin with, but working with what we have, I would rather rename the local workspace to `workspace-configs`, and name the published package `configs`.
I'm also wondering if we need the double package.json here or if it will get us into trouble when we try to publish. We could simplify this a bit by only publishing the ESLint config for now and waiting on prettier, or creating separate packages for ESLint and Prettier configs.

Gatsby has it's own prettier config anyway, and I believe only ESLint for our current needs. I think prettier being published is more of a nice to have at this time.

mitchellmarkoff:
@CobyPear Most of the diff here is renaming import routes to point to the updated local workspace naming, `workspace-configs`. I changed up the approach from before and was able combine ESLint and Prettier into one package this time. I am liking this a lot more. If there are any issues with the single package, Prettier can be removed from this easily and we can go ahead with just ESLint for now.

packages/configs/index.js
```diff
@@ -0,0 +1,28 @@
...
+		'*.test.*',
+		'__tests__/fixtures/**/*',
+		'setupFile.ts',
+		'__mocks__/**/*',
+		'rollup.config.js',
+		'vite.config.*',
+		'jest.config.*',
+		'lint-staged.config.*',
+		'/dist',
+		'/node_modules'
```
CobyPear, line 26: Since a lot of these dirs relate to our npm packages, we can probably clean this up a bit for the published version. We can start with the following dirs:
```js
node_modules
build
static
public
```

packages/configs/index.js
```diff
@@ -0,0 +1,28 @@
...
+	env: {
+		node: true,
+	},
+	plugins: ['@typescript-eslint'],
+	extends: [
+		'prettier',
+		'plugin:prettier/recommended',
+		'eslint:recommended',
+		'plugin:@typescript-eslint/recommended',
+		'plugin:@typescript-eslint/recommended-requiring-type-checking',
```
CobyPear, line 13: Will this break if it tries to lint a project without typescript? Do we need to remove these for this first publish?

packages/configs/package.json
```diff
@@ -0,0 +1,36 @@
...
+		"type": "git",
+		"url": "https://github.com/pantheon-systems/decoupled-kit-js"
+	},
+	"author": "@pantheon-systems",
+	"files": [
+		"./index.js",
+		"./react.js",
+		"./prettier.config.js"
+	],
+	"main": "./index.js",
```
CobyPear, line 18: This looks correct according to the docs, but I'm worried that something about this won't work because the name of the module is not prefixed with "eslint-config". Did you try using `npm link` to test this locally outside of the monorepo?

See https://eslint.org/docs/latest/extend/shareable-configs#creating-a-shareable-config for more info.

packages/configs/README.md
```diff
@@ -0,0 +1,16 @@
+# @pantheon-systems/configs
+
+ESLint configs for `@pantheon-systems` npm packages.
+
+The configs are currently unpublished.
```
CobyPear, line 5: This section could be more accurate.

packages/configs/README.md
```diff
@@ -0,0 +1,16 @@
...
+## How To Add A New ESLint config
+
+1. Add a \*.js file to the root of this directory with the name of the config,
+   for example `vue.js`
+1. Extend the base plugin from your plugin unless it is meant to be standalone
+1. If a new parser is used, add it to the `dependencies` in the `package.json`
+1. If new plugins are used, add them to the `peerDependencies` in the
+   `package.json`
+1. From your .js file, export your config
+1. Add the config to the `files` array in the `package.json`
```
CobyPear, line 16: This is probably still fine but ensure the accuracy here. We might want to move this to CONTRIBUTING or keep it in the local one? Unsure of how detailed this readme needs to be at this point.

.changeset/little-pillows-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pantheon-systems/wordpress-kit": patch
```
CobyPear, line 2: This is actually a case where a single changeset for all of these might make more sense xD. No need to change it right now, but whenever the change is the same across multiple packages, I think in that case a single changeset is fine. We should probably come up with some guidelines around the changesets at some point.

CobyPear (changes requested):
A few questions/comments– I'm a bit concerned about the ts plugin being used on a non-ts project. the reorganization looks good tho!

packages/configs/index.js
mitchellmarkoff: Seems to still work in projects without typescript. I used `npm link` to bring this package into a new create react app project outside of the monorepo and experienced no issues.

CobyPear: Great, thank you for confirming

packages/configs/README.md
mitchellmarkoff: After thinking about it more I feel like making this README light and keeping a more in depth one with contributing details in our local workspace makes sense. Just pushed an update to reflect this.

.changeset/little-pillows-tell.md
mitchellmarkoff: That makes sense. I think eventually having some official changeset guidelines would be helpful and worth doing😆

packages/nextjs-kit/.eslintrc
```diff
@@ -1,6 +1,6 @@
 {
-"extends": "@pantheon-systems/eslint-config/react",
-  "parserOptions": {
-    "project": ["./tsconfig.json"]
-  }
+	"extends": "@pantheon-systems/eslint-config-dk/react",
```
CobyPear, line 2: For the packages, I think we will want to continue using the workspace configs

packages/configs/package.json
```diff
@@ -20,9 +20,6 @@
 		"@typescript-eslint/parser": "^5.49.0"
 	},
 	"prettier": "./prettier.config.js",
```
CobyPear, line 22: Because we can't export the prettier config from the same package as the ESLint config (or if we can it's confusing because of the name of the package), let's move the prettier config to its own package.

CobyPear (changes requested):
I still don't see the package `name` changed to the correct format. I still see `@pantheon-systems/configs`. I would expect `@pantheon-systems/eslint-config-dk` (if we're going with `dk`).

mitchellmarkoff:
> I still don't see the package `name` changed to the correct format. I still see `@pantheon-systems/configs`. I would expect `@pantheon-systems/eslint-config-dk` (if we're going with `dk`).

My bad I must have reverted that file since I had thought I pushed the name change, updated now.

CobyPear (commented):
@mitchellmarkoff I just pushed up some changes:

- use `decoupled-kit` instead of `dk`. I don't mind the shorter name but after consulting with Brian I agree that the full name will be less ambiguous.
- Made the 'prettier' config package more generic. We should be able to export any future configs from here (tsconfig for example). So, instead of changing this later I am rename this to `decoupled-kit-configs`
- Add our tooling (lint-staged, prettier, eslint). Yes, even our configs need configs :)
- Remove the 'dependencies' - ESLint docs say to mark everything as peer. Our workspace package is a bit different because it's a workspace and is not published, so it doesn't follow the same rules.
- Revert change to nextjs-kit: it was still using the new config but I want our monorepo to continue using workspace configs. These published configs are for the generated starters.
- Updated some README language
- Add `browser: true` to the env field in the react config. Without this, linting a gatsby project was throwing some unexpected linting errors like "window is not defined"

Let me know if this makes sense.

mitchellmarkoff (commented):
@CobyPear Thanks for that outline, the changes look good and make sense! I will be bookmarking this PR to refer back to whenever the time comes to edit/add packages. A lot of good information is in here that is new to me. Looks good to merge. I was able to successfully test the package locally on projects both in and outside of the monorepo, TS and non TS.

CobyPear (approved):

PR #547, opened 2023-02-24
DB-5507-DB-5509: Refactor to remove node-plop

CobyPear:
## What changes were made?
- Refactor functions using node-plop
	- Add action runner
	- Add handlebars instance
	- Refactor addWithDiff to dedupe colliding templates
- Refactor Generators to use new action runner
- Refactor main() to use new generators and action runner
- Refactor unit tests
- Add unit tests where needed

functions in `dedupeTemplates` did not receive unit tests at this time.
Most functions should have TSDoc comments.
API reference was not regenerated.

The CLI should function the same as it did. Note the console output is more enhanced.
Open to any and all feedback here including "we should stick with `node-plop`" 😅 

Unit tests refactored to reflect changes as well as testing the actual CLI locally.

## Additional information
Sorry for the large diff


packages/create-pantheon-decoupled-kit/package.json
```diff
@@ -18,7 +18,6 @@
 		"npm init"
 	],
 	"type": "module",
-	"main": "/dist/bin.js",
```
CobyPear, line 21: This was causing `pnpm link` to fail (and also it comes up as a warning in publint.dev)

packages/create-pantheon-decoupled-kit/src/utils/dedupeTemplates.ts
```diff
@@ -0,0 +1,91 @@
...
+			delete mergedPaths[template];
+			mergedPaths[template] = { base, addon };
+		}
+	}
+	return mergedPaths;
+};
+
+/**
+ * In case there is a template path that exists in two or more generators,
+ * we want to favor the addon. This way, we avoid writing to a
```
backlineint, line 66: I'd agree that favoring addons makes sense, but I do have some questions about the details. Some of this is probably me not being fully in the loop on how addons work.

* If a base generator and and addon generator both operate on recipes.js, does only the addon template end up being added in this case? If that is true, I'm not sure it is 'bad', but it does change some of my assumptions about how addons would have to be structured.
* What happens if there are multiple addons that operate on the same file? Does the last addon win?

backlineint (commented):
Still have some manual testing to do, but submitting a review in the meantime. Didn't see anything that I think must be changed, but did have a few questions.

packages/create-pantheon-decoupled-kit/src/types.ts
```diff
@@ -8,8 +9,91 @@ declare module 'vitest' {
...
+	[key: string]: DataMember | Record<string, DataRecord | DataMember>;
+};
+
+// TYPE PREDICATES
+
+/**
+ * @param arg a variable
+ * @returns true if the variable is a string, false otherwise
+ */
+export const isString = (arg: unknown): arg is string => {
```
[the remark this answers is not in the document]
CobyPear: > This feels too primitive and not something we'd need to create our own type predicate for. 

Yeah, this may be a hasty abstraction. 

> Less verbose?

yes slightly. I'm ok with removing it.

packages/create-pantheon-decoupled-kit/src/utils/dedupeTemplates.ts
CobyPear: > If a base generator and and addon generator both operate on recipes.js, does only the addon template end up being added in this case? 

Yes that is the case here. It's possible that we won't need to do this if we structure add-ons differently, but it seemed like the path of least resistance for now

> If that is true, I'm not sure it is 'bad', but it does change some of my assumptions about how addons would have to be structured.

I'm open to restructuring this if there's another pattern that makes more sense. I'm hoping that moving more things to packages and using more partials will allow us to avoid clashing template files in the first place. I think there's been a bit of a learning curve to use partials so we aren't using them as much as I would like at the moment but they may help solve some things.

> What happens if there are multiple addons that operate on the same file? Does the last addon win?
currently that is the case. We could use a number instead of a boolean for the `addon` property indicating it's priority (lower wins) or something like that in case we want more fine grained control over that, but for now the last called generator wins.

backlineint: Gotcha. I could imagine cases where this behavior could become a problem, but I'm just making stuff up :) Knowing the bounds of this helps, and if this behavior becomes an issue in the future, we can tackle it with a real-world use case.

PR #552, opened 2023-02-27
Fix missing dependency

CobyPear:
## What changes were made?
Handlebars moved to prod dependencies, which is the correct place 😅 

Will push another canary release after this is merged to confirm the fix is working as intended.
## Additional information


mitchellmarkoff (approved):
🕺

PR #569, opened 2023-03-06
DB-5495: Add Published `eslint` Configs to Gatsby Generator

mitchellmarkoff:
## What changes were made?
- When creating a gatsby-wordpress project using create-pantheon-decoupled-kit, it will be successfully linted at the end of the project creation process
- The Gatsby template’s package.json includes our new lint configs

A new gatsby-wp starter generated through the CLI was observed to lint after the creation process and support the `pnpm lint` command.

## Additional information


packages/create-pantheon-decoupled-kit/package.json
```diff
@@ -43,8 +43,10 @@
 		"lint-staged": "lint-staged"
 	},
 	"devDependencies": {
-		"@pantheon-systems/workspace-configs": "*",
+		"@pantheon-systems/decoupled-kit-configs": "0.1.0-canary.0",
 		"@pantheon-systems/eslint-config": "*",
+		"@pantheon-systems/eslint-config-decoupled-kit": "0.1.0-canary.0",
```
CobyPear, line 48: I wouldn't expect this `package.json` to change – why do we need the non-workspace configs added here?

pnpm-lock.yaml
```diff
@@ -3028,7 +3032,7 @@ packages:
     peerDependencies:
       react: '*'
     dependencies:
-      '@types/react': 18.0.27
+      '@types/react': 17.0.40
```
CobyPear, line 3035: I wouldn't expect this change here unless I'm missing something.

packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/package.json.hbs
```diff
@@ -37,14 +39,24 @@
 	},
 	"devDependencies": {
 		"@babel/core": "^7.20.12",
+		"@pantheon-systems/decoupled-kit-configs": "0.1.0-canary.0",
+		"@pantheon-systems/eslint-config-decoupled-kit": "0.1.0-canary.0",
```
CobyPear, line 43: This is fine but will need to be changed once the non-canary version is released.. We can use `"latest"` here for the version to get around this.

packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
+	"extends": "@pantheon-systems/eslint-config-decoupled-kit",
```
CobyPear, line 2: I would expect the react config here. Also you can omit the `eslint-config` prefix here if you'd like
```suggestion
	"extends": "@pantheon-systems/decoupled-kit/react",
```

CobyPear: This will probably cause a bunch of 'prop-types' errors. I believe this is because Gatsby ignores this rule or handles it somehow under the hood. I think for now it's worth ignoring this rule. When we figure out our TypeScript story for the starters, we can revisit this.

CobyPear (changes requested):
There are a few changes here I was not expecting, but it's close!

packages/create-pantheon-decoupled-kit/package.json
mitchellmarkoff: @CobyPear Yeah this was interesting, it is not required at all for the functionality of this feature but the monorepo linter was yelling and preventing me from committing without it there.
Screen Shot 2023-03-06 at 2 08 03 PM

CobyPear: `lint-staged` worked for me on this branch without those dependencies, might be some config file in your tree that is looking for the published config?

CobyPear: Oh yeah, if you look at the path in the error, it is looking at the config file in the template for some reason.

CobyPear (approved):
Approved! I'm holding off on merging until we get CI sorted out.

CobyPear:
CI is mostly sorted, merging this!

PR #572, opened 2023-03-07
DB-5253: Fix for conditional action

CobyPear:
## What changes were made?
Hopefully this fixes the GH Action conditional steps so that the docs site still works.

This should also be a good test of our most recent canary release of the CLI.
## Additional information

### DIFF
--- a/.github/workflows/canary-sites-split.yml
+++ b/.github/workflows/canary-sites-split.yml
@@ -52,25 +52,30 @@ jobs:
           # canary docs site
           - local_path: 'web'
             split_repository: 'decoupled-kit-docs-canary'
+            generator_cmd: ''
 
     steps:
-      - uses: actions/setup-node@v3
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Generate starter
+        if: ${{ matrix.generator_cmd != '' }}
+        uses: actions/setup-node@v3
         with:
           node-version: 18
-      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/canary-sites-split.yml
```diff
@@ -52,25 +52,30 @@ jobs:
...
         with:
           node-version: 18
-      - uses: actions/checkout@v3
-        if: ${{ matrix.generator_cmd }}
         # install canary version of the CLI
-      - run: npm i --global create-pantheon-decoupled-kit@canary
-        if: ${{ matrix.generator_cmd }}
         # run the command to generate the starter in the matrix.local_path directory
       - run: |
+          npm i --global create-pantheon-decoupled-kit@canary
```
backlineint, line 69: @CobyPear I didn't think this step all the way through during the last PR. Am I correct that this is using the published version of the package? I wonder if it would be possible for there to be a case where this runs before the current version is fully available on npm and as a result we publish an outdated version of the starter. Or am I jumping to conclusions here.

CobyPear: @backlineint Correct, this would be the published canary version. It would mean we would only get the 'latest' changes to the canary sites when we do a canary release. The hidden issue with that is, the `package.json`s are static and do not receive the changesets bumps like our other starters do now, so using the canary version of the CLI does not spit out a starter with the canary version of the packages as one might expect.

backlineint (approved):

PR #586, opened 2023-03-15
DB-5252: Update READMEs to use `npm init` as the default way to create a project

mitchellmarkoff:
## What changes were made?
Updated `README` files across the monorepo to use `npm init` as the recommended way to create a project.

`next-wordpress-starter`
`next-drupal-starter`
`gatsby-wordpress-starter`

`README` files have been viewed online and seen to have working links and correct content.

## Additional information


mitchellmarkoff:
@CobyPear  LMK if these new instructions should be more verbose or not, I know the goal was to lean on the dev docs for a lot of the info.

starters/gatsby-wordpress-starter/README.md
backlineint, line 1: Probably fine either way for now, but I wouldn't expect us to be making manual updates to the starters in the codebase anymore. I'm actually assuming we'll delete these once we're confident the new process is working.

backlineint (changes requested):
I was assuming that pages in the docs like: https://decoupledkit.pantheon.io/docs/frontend-starters/nextjs/nextjs-drupal/introduction would also be updated to suggest using the CLI instead of create-next-app. @CobyPear is this in line with what you were expecting?

starters/gatsby-wordpress-starter/README.md
CobyPear: >I was assuming that pages in the docs like: https://decoupledkit.pantheon.io/docs/frontend-starters/nextjs/nextjs-drupal/introduction would also be updated to suggest using the CLI instead of create-next-app. @CobyPear is this in line with what you were expecting?

Yes, anywhere that we currently ask the user to use some other CLI command. Not sure if we should still advertise `git clone` as an option anymore either. We could tho.

>I wouldn't expect us to be making manual updates to the starters in the codebase anymore

Agreed, I think we can safely remove them now given the new CI process seems to have worked for the latest release.

mitchellmarkoff: Ah I had missed the mentions in the docs site itself, will go back and update that.

> Agreed, I think we can safely remove them now given the new CI process seems to have worked for the latest release.

@CobyPear Will removing the starters come in its own ticket or should they be pulled out here?

CobyPear (changes requested):
I only left comments on the one doc but they will apply to all starters.

web/docs/Frontend Starters/Next.js/Next.js + WordPress/setting-environment-variables.md
```diff
@@ -11,10 +11,11 @@ In order to fetch data from the WordPress instance, Next.js needs to know the
 endpoint at build time. For local development, the starter kit uses
 [dotenv](https://www.npmjs.com/package/dotenv).
 
-When you clone your decoupled frontend repo, create a `.env.development.local`
-file. In this file, update the `WPGRAPHQL_URL` and `IMAGE_DOMAIN` with your
-WordPress GraphQL endpoint, and the `IMAGE_DOMAIN`. If the `WPGRAPHQL_URL` and
-`IMAGE_DOMAIN` are the same, you can omit setting the `IMAGE_DOMAIN`.
+When you generate your decoupled frontend repo, create a
+`.env.development.local` file if one does not already exist. In this file,
```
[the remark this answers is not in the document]
mitchellmarkoff: @CobyPear I had chose this wording since as it is now we still are advertising `git clone` as an (not preferred) option so in a case where a user goes that route they would still need to create `.env.development.local`. I am open to changing this to assume they used a generator if we don't think that assumption will lead to confusion.

> Maybe we could add in a note that the WPGRAPHQL_URL/BACKEND_URL can be set as a argument when running the CLI?

I think at this point in the docs the user would have likely already generated a project, maybe that would be more suited in the CLI docs page? Could see and argument to have a note in both places as well.

starters/gatsby-wordpress-starter/README.md
CobyPear, line 1: Instead of updating this file with new info, maybe it's better to add a deprecation message and point people to the dev docs on the CLI instead. That way we can signal these starters aren't getting updates and the CLI should be used to get the latest starters. @backlineint how do you feel about replacing the current README content of each starter with a deprecation message?

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/intro.md
```diff
@@ -26,36 +26,34 @@ backend.
...
 
 There are two methods to creating a new project based on the
-`gatsby-wordpress-starter`:`
+`gatsby-wordpress-starter`:
 
-1. Clone the starter repo directly
-2. Use the `gatsby-cli`
+1. **Preferred Method** - Using our `create-pantheon-decoupled-kit`, or the
+   "CLI"
+2. Clone the starter repo directly
```
CobyPear, line 33: I'm still not sold on this method being advertised at all.

CobyPear (changes requested):
A few thoughts: maybe replace the starter's READMEs with a deprecation message?
Also, still not completely convinced we should advertise the git clone method.

starters/gatsby-wordpress-starter/README.md
backlineint: @CobyPear and @mitchellmarkoff that sounds like a good idea to me.

web/docs/Frontend Starters/Gatsby/Gatsby + WordPress/intro.md
backlineint: Can kind of see both arguments here, but the thing that tips me over the edge is that there are actually kind of three methods - the CLI, cloning the repo directly, and selecting the project in the dashboard (which technically clones the repo behind the scenes.) Also thinking about now Next, Astro etc present this, they rarely if ever point people to the repo directly.

So I'll side with only suggesting the CLI as well.

PR #589, opened 2023-03-17
DB-5055: tailwindcss addon

CobyPear:
## What changes were made?
- Added a new action, `addDependencies`, that will add any dependencies listed in a generators `data` field to the `package.json` found in the `outDir`.
- Added a new generator `tailwindcss-addon` that uses the `addDependencies` action, as well as the `addWithDiff` action to bootstrap a project with tailwindcss dependencies and configs
- Added a `nextSteps` option to generators, so that any crucial information relating specifically to the generator can be logged to the console when all actions are complete.
- Added tests for the `addDependencies` action
-Some tweaks to the `Input` type
- Added this add-on to the CLI docs.

cli, docs
wrote tests, and used the generator on a React project created with `npm init vite`. After minimal changes (importing the proper css file and removing styles that were already there), I was able to use tailwindcss with no issues.
## Additional information

Regarding the 'nextSteps' for this addon, I am open to suggestions here on what the actual info should be for the nextSteps, as we can't know everything about the user's project to make this no config unless we add more prompts to this generator. Which maybe is what we want to do? But I think I'd rather keep the prompts for this one minimal.

And regarding the docs:  Should we have a new section for 'generic' add-ons? Or is it ok to add these under the `Add-ons` section of our CLI doc (which is  what I did here)

### DIFF
--- a/web/docs/Frontend Starters/create-pantheon-decoupled-kit.md
+++ b/web/docs/Frontend Starters/create-pantheon-decoupled-kit.md
@@ -61,13 +61,23 @@ results. These `project` generators should be used only in conjunction with
 
 ### Add-ons
 
+There are some utility add-ons that are available to any project. These add-ons
+are:
+
+- tailwindcss-addon
+
+  This add-on will bootstrap a project with dependencies and config files
+  necessary to start using [tailwindcss](https://tailwindcss.com). Minimal
+  additional setup may be required, such as importing the css file or changing
+  the paths in the `tailwind.config.cjs`
+
 Available add-ons per project generator are as follows:
 
 | Project Generator | Available Add-ons       |
 | ----------------- | ----------------------- |
 | next-drupal       | next-drupal-umami-addon |
 | next-wp           | next-wp-acf-addon       |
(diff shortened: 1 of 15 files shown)


web/docs/Frontend Starters/create-pantheon-decoupled-kit.md
backlineint, line 1: Something to revisit with the follow up tickets, but I wonder if at least a reference to this addon in the individual starter addon docs also makes sense.

backlineint (approved):
One issue I noticed while running this with a related generator is that we end up with duplicate tailwind / postcss config files with different file extensions (cjs/js) I'm assuming the next ticket to de-tailwind the starters will address this.

I was also curious to see what happened if you used this addon with a project that already had some of the necessary dependencies, but specify a different constraint. The entry in package.json is replaced with the one in the addon, including the new version constraint. At the moment I think this makes sense, and we're clear in our output about what we're adding to package.json. Possibly something to consider though if we have a future addon with more complicated or common dependencies.

PR #592, opened 2023-03-23
DB-5227: tailwindless next-wp (part 1)

CobyPear:
## What changes were made?
- Add css modules to replace tailwind styles
- Add "Would you like tailwindcss" prompt to `next-wp`
(I just realized this question should probs also be on the `next-wp-acf-addon` generator? otherwise it will always be tw by default unless you use this add-on while creating a project with no tw.)
- Consolidate `next-wp-acf-addon` templates into `next-wp` templates. Using handlebars conditionals here instead.
- Add `tailwindless-next` shared templates
- Add `tailwind-shared` shared templates
- Add a partial for tailwindcss dependencies
- Export `PostGrid` and `PageGrid` from `components/grid.jsx` instead of creating them in each page they are needed

`cli` > `templates` > `next-wp`, `next-wp-acf-addon`, `tailwindless-next`, `tailwind-shared`
Some slight modification to `main()` to pass through tailwindcss args
Slight modifications to effected generators

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/next-wp/pages/posts/[...slug].jsx.hbs
```diff
@@ -3,22 +3,22 @@ import { ContentWithImage } from '@pantheon-systems/nextjs-kit';
...
-import { withGrid } from '@pantheon-systems/nextjs-kit';
-
 import { getFooterMenu } from '../../lib/Menus';
+import { IMAGE_URL } from '../../lib/constants';
 import { getPostByUri, getPostPreview } from '../../lib/Posts';
+{{#if wpAcfAddon}}
+import { PostGrid } from '../../components/grid';
+{{/if}}
+{{#unless tailwindcss}}
+import styles from './[...slug].module.css';
```
Comment (backlineint) [inline packages/create-pantheon-decoupled-kit/src/templates/next-wp/pages/posts/[...slug].jsx.hbs:13]: I think wires are crossed somehow here. When I create a starter kit, viewing posts fail because this css module doesn't exist. The actual hbs template seems to only apply if the acf addon is being used. So maybe the conditional is wrong here?

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/components/grid.module.css
```diff
@@ -0,0 +1,34 @@
...
+  height: inherit;
+}
+
+.card {
+  cursor: pointer;
+  border-radius: var(--2);
+  border: var(--0_5) var(--light-gray) solid;
+  overflow-x: hidden;
+  /* 
+   * generated by https://tailwind-to-css.vercel.app/
```
backlineint, line 13: Cool / slightly horrifying that this exists :)

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/components/layout.module.css
```diff
@@ -0,0 +1,26 @@
...
+.footerCopy {
+  margin: var(--x-auto);
+}
+
+.footerCopy > [href] {
+  text-decoration: underline;
+}
+
+.footerCopy > [href$='nextjs.org/'] {
+  color: var(--wite);
```
backlineint, line 21: Typo for `white` here?

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/components/layout.module.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.footerCopy > [href$='wordpress.com/'] {
+  color: var(--blue);
```
backlineint, line 25: Technically this seems to be using different blue than the WP branded blue used by the tailwind version.

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/components/layout.module.css
```diff
@@ -0,0 +1,26 @@
...
+
+.layout > main {
+  margin-bottom: auto;
+}
+
+.footerCopy {
+  margin: var(--x-auto);
+}
+
+.footerCopy > [href] {
```
backlineint, line 16: Targeting the href attribute for something like this is a little unusual. Any reason this isn't `a`? Same comment goes for similar selectors throughout.

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/pages/examples/auth-api.module.css
```diff
@@ -0,0 +1,35 @@
...
+.container > div {
+  margin: var(--x-auto);
+}
+
+.container > div > p {
+  margin-top: var(--12);
+  font-size: var(--5);
+}
+
+@media (max-width: 1024px) {
```
backlineint, line 27: Should this be one of the screens variables? Or perhaps I'm vaguely remembering that you can't use css custom properties in media queries for some reason.

backlineint (changes requested):
In addition to the inline comments, a few styling differences that I think are worth addressing:

* On pages and posts the body text is too wide in comparison to the tailwind version. Can be seen pretty easily on /pages/sample-page
* Subtle, but the font color for the body text on the examples pages is darker than the font color on the related tailwind examples.

Looking into the font color differences above I also noticed that I was still seeing some tailwind classes, and then eventually tailwind styles. I think this is a result of next-js kit, which is still bringing in Tailwind as a dependency. It seems like maybe the components should move into CLI templates as a way around this? This seems like something that should be done in a follow up ticket - I'll create one.

Also, what does all of this mean for the block editor plugins? Does that also have the result of pulling in tailwind as a dependency? Or does it compile things down to css?

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/pages/examples/auth-api.module.css
CobyPear: > you can't use css custom properties in media queries for some reason.

Yeah this is the reason :(
I am missing the comment though, I'll add that.

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/components/grid.module.css
CobyPear: It wasn't as useful as I thought it would be, but it's nice for one-offs like this.

packages/create-pantheon-decoupled-kit/src/templates/next-wp/pages/posts/[...slug].jsx.hbs
Comment (CobyPear) [reply packages/create-pantheon-decoupled-kit/src/templates/next-wp/pages/posts/[...slug].jsx.hbs:13]: Looks like the conditional was wrong, but even when it is correct, this issue highlights the fact that this pattern with an empty `templates` array in the generator won't really work for the add-ons. I am breaking out these files into partials.

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/components/layout.module.css
CobyPear: No good reason, replacing with `a` :)

CobyPear:
> I think this is a result of next-js kit, which is still bringing in Tailwind as a dependency. It seems like maybe the components should move into CLI templates as a way around this?

`nextjs-kit` brings it's own stylesheet with prefixed tailwind classes. This stylesheet needs to be imported into the `app.tsx` or the styles aren't applied, but the components do have those class names regardless of the stylesheet. So, tailwind does not need to be a dep for those styles, but it also shouldn't hurt that those classes are there unless they are really interfering with some other code somehow.

> Also, what does all of this mean for the block editor plugins? Does that also have the result of pulling in tailwind as a dependency? Or does it compile things down to css?

 The plugin requires tailwind, but you could still use vanilla css with it configured.

backlineint (approved):

PR #593, opened 2023-03-24
DB-5715: Add Placeholder Search Route + App Layout

mitchellmarkoff:
## What changes were made?
- A root layout was added to the `/app` directory
- `/search` page created, displays only the heading ‘Search Results’
- Submitting a search redirects to `/search?=<term>`

`cli`

Generated a new project using the locally linked `create-pantheon-decoupled-kit package`. The `next-drupal` and `next-drupal-search-api-addon` generators were used along side each other with no problems. A new project was generated with the new search functionality working as expected.
## Additional information
**A few things to note**
- There are two versions of `search-input.jsx` that exist because Next 13 uses [its own version of `useRouter`](https://beta.nextjs.org/docs/routing/linking-and-navigating#userouter-hook) that is under `next/navigation`. Is it not backwards compatible with pages outside of the app directory so two components exist, one using `next/router` the other using `next/navigation` 
- A lot of issues came up with this ticket in trying to use the optional catch all routing design as originally planned. Essentially, our [[...alias].jsx](https://github.com/mitchellmarkoff/decoupled-kit-js/blob/canary/packages/create-pantheon-decoupled-kit/src/templates/next-drupal/pages/%5B...alias%5D.jsx) file in the pages directory was hijacking any `router.push()` events trying to reach a catch all route in the app directory. I landed on changing our app directory structure as the simplest path forward. To avoid the `[...alias]`  route from getting called, our search now routes `search?=<term>` instead of `search/<term>`.


packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/app/layout.jsx
CobyPear, line 1: Are we going to have tailwindless options for addons as well? If not, we should be packaging these components into `nextjs-kit` or even some other package so that the styles can be self contained. If these are meant as examples, especially to be called with (or without) a base generator, I would expect them to work with or without tailwind.

mitchellmarkoff: You bring up a good point here. Is there a route you have chosen for rewriting any add-ons in your current tailwind work? Creating a `nextjs-drupal-kit` would have the strength of self contained styles but I wonder if separating out these components, which are the bulk of some add-ons, would make it harder for a user to dissect our examples.

cc: @backlineint any thoughts on this?

CobyPear: If the add-on is simple enough I think it could be included in the base template as a conditional. That is what I did for the `next-wp-acf-addon` in the tailwindless rewrite. There is still a [single css file needed for the tailwindless version of that, which is part of the `tailwindless-next` templates](https://github.com/pantheon-systems/decoupled-kit-js/pull/592/files#diff-63dc4ec0e1c71ebe38a8c3d596da478ed133004d4a77fea6f0284390f6b4b2f2).

For this you can use similar `if` and `unless` handlebars blocks to render the correct version based on the input. I also added a small bit of logic to `addWithDiff` so it would skip rendering handlebars files that rendered out an empty string. This way, you could have entire files that are only rendered in certain cases.

CobyPear: Also I think this file could make use of partials since it's mostly the same as our common `layout.tsx`. If you want to leave some of this stuff how it is, We can pair up when I do part 2 of the tailwindless stuff and we can go through 'de-composing' and 're-composing' some of these templates

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/tailwind.config.js
CobyPear, line 1: I think this file will also be handled by the shared-tailwind templates and we can refactor when we do `next-drupal` tailwindless

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/app/search-input.jsx
```diff
@@ -0,0 +1,39 @@
...
+	const [searchQuery, setSearchQuery] = useState('');
+
+	const router = useRouter();
+
+	const onSearch = (event) => {
+		event.preventDefault();
+
+		if (typeof searchQuery !== 'string') {
+			return;
+		}
```
CobyPear, line 14: This can probably just check for `searchQuery`
```suggestion
		if (!searchQuery) {
			return;
		}
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/app/search-input.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+		const encodedSearchQuery = encodeURI(searchQuery);	
+		router.push(`/search?=${encodedSearchQuery}`);
+	};
+
+	return (
+		<form onSubmit={onSearch}>
+			<input
```
CobyPear, line 22: Form inputs should have a `label` element for a11y reasons. I am not 100%, but we might also want some aria labels and attributes on this markup

CobyPear (changes requested):
Made a few comments re: the form element itself that could be addressed now, but I think some of the other things can wait until the tailwindless refactor of the next-drupal templates.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/next.config.js
CobyPear, line 1: I still think this file doesn't need to be here in its current form. We should be able to use partials and handlebars parameters to share the common next.config code across different nextjs based generators.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/components/search-input.jsx
```diff
@@ -9,30 +9,35 @@ const SearchInput = () => {
...
 		event.preventDefault();
 
-		if (typeof searchQuery !== 'string') {
+		if (!searchQuery) {
 			return;
 		}
 
-		const encodedSearchQuery = encodeURI(searchQuery);
-		router.push(`/search/${encodedSearchQuery}`);
+		const encodedSearchQuery = encodeURI(searchQuery);
```
CobyPear, line 16:
```suggestion
		const encodedSearchQuery = encodeURI(searchQuery);
```
looks like some extra space was added?

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/components/search-input.jsx
```diff
@@ -9,30 +9,35 @@ const SearchInput = () => {
...
+			</label>
+		<button
+			type="submit"
+			id="submit-btn"
+			className="ml-1.5 text-black rounded-full ring-[1px] ring-black px-5 py-1 hover:bg-blue-500"
+			aria-label="Submit Search"
+		>
+			Submit
+		</button>
+	</form>
```
CobyPear, line 40: 👏 yay a11y!

CobyPear (changes requested):

CobyPear:
One more small one: the `/recipes` link still exists on the layout even if the umami demo data add-on is not selected.

mitchellmarkoff:
> One more small one: the `/recipes` link still exists on the layout even if the umami demo data add-on is not selected.

Oo good catch with this one 👍

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-umami-addon/components/layout.jsx
CobyPear, line 1: Is this file necessary?

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/app/layout.jsx
CobyPear: @mitchellmarkoff I think this file might have gotten missed for the partials treatment. LMK if you want to pair up on this one.

CobyPear (changes requested):
Sorry about that recipes thing, Mondays amirte?

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-umami-addon/components/layout.jsx
mitchellmarkoff: Can and will be removed now we are taking out the recipes thing.

mitchellmarkoff:
@CobyPear 

> @mitchellmarkoff I think this file might have gotten missed for the partials treatment. LMK if you want to pair up on this one.

I think I got it this time around, your handlebar playground example cleared some things up for me! If anything still doesn't look right with the partials maybe we could pair in the AM 👍

packages/create-pantheon-decoupled-kit/src/templates/partials/nextDrupalLayout.hbs
```diff
@@ -16,24 +15,23 @@ export default function Layout({ children, footerMenu, preview = false }) {
...
+        {{#if app}}
 		<html lang="en">
 			<body>
+        {{/if}}
 				<div className="min-h-screen max-h-screen min-w-screen max-w-screen flex flex-col overflow-x-hidden">
 					{preview && <PreviewRibbon />}
+                 {{#if appDir}}
 					<div className="ps-flex ps-flex-row ps-flex-wrap sm:ps-flex-nowrap ps-justify-between ps-max-w-screen-md ps-mx-auto">
-						<Header navItems={navItems} />
+                        <Header navItems={navItems} />
```
CobyPear, line 27: The indentation feels a bit off here. It should be level with the div below it.

packages/create-pantheon-decoupled-kit/src/types.ts
```diff
@@ -46,6 +46,10 @@ export interface DecoupledKitGenerator<Prompts extends Answers> {
 	 * Any message(s) to be rendered after actions are successfully completed.
 	 */
 	nextSteps?: string[];
+	/**
+	 * Set to true if the generator is to use the experimental app directory from  Next JS.
+	 */
+	appDir?: boolean;
```
CobyPear, line 52: This doesn't need to be added to the types. Use the `data` field in the generator to pass arbitrary data to the templates.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/components/layout.jsx.hbs
```diff
@@ -0,0 +1 @@
+{{> nextDrupalLayout }}
```
CobyPear, line 1:
```suggestion
{{> nextDrupalLayout}}
```
We haven't really discussed our handlebars style guide, but imo the closing tag should not have a space before it. Open to other ways of doing it though.

CobyPear (changes requested):
Mostly some formatting that eventually should be solved by proper linting of these templates, but until then, unfortunately we need to do some manual formatting.

packages/create-pantheon-decoupled-kit/src/types.ts
mitchellmarkoff: Ah good catch, I had confused this with the `addOn` flag originally and meant to remove this.

packages/create-pantheon-decoupled-kit/src/templates/partials/nextDrupalLayout.hbs
```diff
@@ -54,7 +52,9 @@ export default function Layout({ children, footerMenu, preview = false }) {
 						</span>
 					</Footer>
 				</div>
+        {{#if app}}
 			</body>
 		</html>
+        {{/if}}
```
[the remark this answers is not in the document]
mitchellmarkoff: @CobyPear I like the idea of aligning the opening and closing tags to the the most left block of code within them. I did this for the most recent revisions. I think it looks cleanest to have opening/closing tags aligned with themselves and then be centered around the most left tag in their block. LMK what you think

packages/create-pantheon-decoupled-kit/src/templates/partials/nextDrupalLayout.hbs
```diff
@@ -16,24 +15,23 @@ export default function Layout({ children, footerMenu, preview = false }) {
...
-	const removeRecipes = (value, index) => {
-		if (value.linkText === 'Recipes') {
-			footerMenuItems.splice(index, 1);
-		}
-	};
-	footerMenuItems?.forEach(removeRecipes);
-
 	return (
+        {{#if app}}
 		<html lang="en">
```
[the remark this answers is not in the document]
mitchellmarkoff: I spent a good bit of time on this but sent back the revisions without it for the sake of time. It is possible to determine this lang dynamically but may be outside of the scope of this ticket as locales now work quite differently in the app directory. The `context` parameter that we use for locales in `/pages` no longer supports locale information in `/app`. To handle locales next recommends [creating a middleware](https://beta.nextjs.org/docs/guides/internationalization) and using their specific `lang` file structure.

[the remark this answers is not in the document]
CobyPear: Oh jeez! Thanks much for looking into that. Another thing to add to our `i18n` considerations :)

CobyPear (approved):
LGTM. There is one formatting change I am pushing up, along with an interface for the `data` on the generator since the type refactor was merged in.

On the formatting, I think the partials themselves should have the newline, and when we call the partial it should not have one. That way we only get 1 newline char when we render those templates.

PR #596, opened 2023-03-29
DB-5227: Fix missing dependency

CobyPear:
## What changes were made?
`tailwindcss` is a required dependency for our `wordpress-kit` because of the plugin. This means that the output of a tailwindless next-wp project will fail without this dependency. If `auto-install-peers` is set to true, depending on the package manager, this isn't an issue. But we should not rely on users needing to install this dependency themselves.
Adds `tailwindcss` back to the `next-wp` package.json in all cases

Tested locally with the watch script
## Additional information

### DIFF
--- a/.changeset/chilled-icons-obey.md
+++ b/.changeset/chilled-icons-obey.md
@@ -1,0 +1,6 @@
+---
+'create-pantheon-decoupled-kit': patch
+---
+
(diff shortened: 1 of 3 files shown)


mitchellmarkoff (approved):
LGTM 👍

PR #606, opened 2023-04-03
DB-5807: workspace cleanup

CobyPear:
## What changes were made?

- Remove starters, add starters workspace to gitignore
- Update dependencies
- Update pnpm version
- Update relevant docs
- Remove .npmrc
- Update CI scripts with new pnpm version
- Some dependency updates for the gatsby-starter as well including the fixed `gatsby-source-wordpress` version
- Add changeset
 Bonus scripts:
- generate-starters will generate the now removed starters in the starters workspace
- check-built will only build packages that were not previously built

The `pnpm audit` overrides seem to be working now which brings up an interesting but with the `got` package. The override is removed and added to `ignoreCves`. While the override works for our docs site which is where `got` is being used by `docusaurus`, our gatsby starter also uses `got` but can not support the latest version (latest version is ESM, gatsby 4 doesn't support ESM).

starters workspace, packages, monorepo, CI scripts
used the new `generate-starters` script and checked each starter with the workspace packages.
CI needs to be verified.
## Additional information

@backlineint  @mitchellmarkoff Given we are removing the starters and will no longer be running tests for them in CI, we could use `generate-starters` in CI so that we can continue running the snapshot tests. This might require a bit more work so I would suggest a new ticket but I am curious about your thoughts on this.

Also regarding the changesets, I was hoping the "fixed" or "linked" setting would make sense for the CLI so that it would be bumped each time a package is bumped. I'm not sure if this makes sense at the end of the day, and we may need to manually make sure the CLI gets at least a patch bump if any of the packages change versions.


backlineint:
> @backlineint @mitchellmarkoff Given we are removing the starters and will no longer be running tests for them in CI, we could use `generate-starters` in CI so that we can continue running the snapshot tests. This might require a bit more work so I would suggest a new ticket but I am curious about your thoughts on this.

A new ticket makes sense. Can see how generate-starters could help cover this. We might even be able to determine which starters changed an only run those snapshots. But maybe not worth the effort until the snapshots are time consuming.

backlineint:
> > @backlineint @mitchellmarkoff Given we are removing the starters and will no longer be running tests for them in CI, we could use `generate-starters` in CI so that we can continue running the snapshot tests. This might require a bit more work so I would suggest a new ticket but I am curious about your thoughts on this.
> 
> A new ticket makes sense. Can see how generate-starters could help cover this. We might even be able to determine which starters changed an only run those snapshots. But maybe not worth the effort until the snapshots are time consuming.

Stubbed out https://getpantheon.atlassian.net/browse/DB-5870 for this.

PR #607, opened 2023-04-03
DB-5717: Implement `getDrupalSearchResults` Function

mitchellmarkoff:
## What changes were made?

- Added a standalone function to the `next-druapl` starter that executes searches based off of user input and returns any results.
- Added unit tests for this new function.
- The new function bubbles up `Surrogate-Key` data to the front end search route.

Locally observed search queries returning expected data against specified index's. Query responses seen to have correctly modified headers.

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/lib/getDrupalSearchResults.js
```diff
@@ -0,0 +1,23 @@
+import { defaultFetch } from '@pantheon-systems/drupal-kit';
+import { ServerResponse } from 'http';
+
+import { NextResponse } from 'next/server';
+
+export const getDrupalSearchResults = async (id, index) => {
```
CobyPear, line 6: I would expect this to live in `drupal-kit` then imported to this add-on and implemented. That way the `getDrupalSearchResults` could be framework agnostic and be potentially reused later.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/app/search/page.js
```diff
@@ -1,4 +1,11 @@
-export default function SearchPage() {
+import { getDrupalSearchResults } from '../../lib/getDrupalSearchResults';
+
+export default async function SearchPage({ searchParams }) {
+	let data = null;
```
backlineint, line 4: Are there any ways we could handle a null response without using let here?

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/lib/getDrupalSearchResults.js
```diff
@@ -0,0 +1,23 @@
...
+	if (!index) {
+		index = 'articles_index';
+	}
+	const newResponse = new ServerResponse(new NextResponse());
+
+	const res = await defaultFetch(
+		`https://search-api-decoupled-drupal-qa.pantheonsite.io/en/jsonapi/index/${index}?filter[fulltext]=${id}`,
```
backlineint, line 13: The API endpoint can't be hardcoded here.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/lib/getDrupalSearchResults.js
mitchellmarkoff: Thanks for breaking apart the different pieces here. With everything you mentioned above, I am also lean towards keeping it where it is with being open to refactoring as the project develops.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/lib/getDrupalSearchResults.js
```diff
@@ -0,0 +1,30 @@
...
+ * @description Helper function to query the Drupal Search API.
+ * @param {import('next').GetServerSidePropsContext} context - Nextjs getServerSideProps context.
+ * @param {string} index - The specific Drupal Search index to be used in the query. Default is an example 'articles_index'.
+ * @see https://www.drupal.org/docs/contributed-modules/search-api for more information about the Drupal Search API.
+ * @returns An array of search results matching the users search term.
+ */
+
+export const getDrupalSearchResults = async (
+	context,
+	index = 'articles_index',
```
CobyPear, line 13:
```suggestion
       {
         query,
         locale,
         res
         index = 'articles_index',
       }
```
I think I favor explicit params here. I see we have some functions that take just `context` but I wonder if those should be updated at some point as well.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/__tests__/getDrupalSearchResults.test.js
CobyPear, line 1: I'm kind of apprehensive to add so many test files to an add-on. This is making me feel more strongly towards breaking this out into `drupal-kit` especially since it feels vanilla enough now that we are no longer using the `/app` router.

CobyPear: This also brings up another topic for "What defines a default": Should add-ons have snapshot tests? I do think this kind of thing should be unit tested, but if it can be unit tested it should probably live in a package, not templates.

CobyPear (changes requested):
The `nextDrupalLayout.hbs` still has some app dir references. I'm also more convinced now that this function should live in `drupal-kit` so that we don't have to ship extra templates in the form of tests.

mitchellmarkoff:
> I'm also more convinced now that this function should live in `drupal-kit` so that we don't have to ship extra templates in the form of tests.

Yeah I see your case for this. Now that we are leaning into `/pages`, this function is a lot more flexible and less Next specific.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/lib/getDrupalSearchResults.js
CobyPear: This doesn't seem to have been addressed– The ask here is to use named arguments. It's probably cleanest to use an interface

```ts
interface GetDrupalSearchResultsParams {
	apiUrl: string;
	locale: string;
	query: string;
	response: ServerResponse;
	index: string;
}

//...tsdoc stuff
const getDrupalSearchResults = async ({
	apiUrl,
	locale,
	query,
	response,
	index = 'articles_index',
}: GetDrupalSearchResultsParams) => {

File: packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx [hunk 5]
```diff
@@ -0,0 +1,92 @@
...
+import { getDrupalSearchResults } from '@pantheon-systems/drupal-kit';
+import Layout from '../../components/layout';
+import PageHeader from '../../components/page-header';
+
+export default function SearchPage({ hrefLang, footerMenu, errorMessage }) {
+	return (
+		<Layout footerMenu={footerMenu}>
+			<NextSeo
+				title="Decoupled Next Drupal Demo"
+				description="Generated by create next app."
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:16]:
```suggestion
				description="Generated by create-pantheon-decoupled-kit."
```
:)

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx
```diff
@@ -0,0 +1,92 @@
...
+		});
+
+		const searchResults = alias
+			? await getDrupalSearchResults(
+					process.env.BACKEND_URL,
+					locale,
+					alias[0],
+					res,
+			  )
+			: undefined;
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:67]: **being picky here and this doesn't have to change**: I think I prefer `null` to undefined in general but I'm not sure it really matters!

packages/drupal-kit/src/lib/getDrupalSearchResults.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @see https://www.drupal.org/docs/contributed-modules/search-api for more information about the Drupal Search API.
+ * @returns An array of search results matching the users search term.
+ */
+
+const getDrupalSearchResults = async (
+	apiUrl: string,
+	locale: string,
+	query: string,
+	response: ServerResponse,
+	index: string | string = 'articles_index',
```
CobyPear, line 21: Use named params here when there are >2.
https://github.com/pantheon-systems/decoupled-kit-js/pull/607#discussion_r1159865364

.changeset/red-crabs-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-pantheon-decoupled-kit": minor
```
CobyPear, line 2: I think the `create-pantheon-decoupled-kit` changeset is a patch version? maybe not. But either way there should be a minor version bump to `drupal-kit`.

CobyPear (changes requested):

CobyPear (approved):
I made some tweaks to the TSDoc comment - TSDoc doesn't require the type annotation like JSDoc does– it's kind of confusing. We will be able to standardize on TSDoc soon enough at least!

PR #619, opened 2023-04-07
DB-5716: Search Results Route

mitchellmarkoff:
## What changes were made?
- Submitting a search in the search box will provide results of articles sourced from Drupal in the main content area of the page.

Search page tested locally and seen to show a functioning article grid based off of search results.

## Additional information


CobyPear (changes requested):
Mostly small code style stuff

CobyPear (changes requested):
I think all this needs is some sort of message when there are no results found, otherwise it feels sort of broken.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx
```diff
@@ -71,12 +91,34 @@ export async function getServerSideProps(context) {
...
+		const [searchTerm] = alias ? alias : [null];
+		const searchResults = searchTerm
+			? (
+					await getDrupalSearchResults({
+						apiUrl: process.env.BACKEND_URL,
+						locale: locale,
+						query: searchTerm,
+						response: res,
+					})
+			  ).data.map((value) => {
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:103]: This map means that the first half of this ternary will always be true, even if it is an empty array. Not sure the ternary is necessary.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx
```diff
@@ -18,11 +26,35 @@ export default function SearchPage({ hrefLang, footerMenu, errorMessage }) {
 			/>{' '}
 			<PageHeader title="Search Results" />
 			<div className="mt-12 mx-auto max-w-[50vw]">
-				{errorMessage ? (
-					<p className="text-xl text-center">
-						⚠️Unable to fetch your search results⚠️
-					</p>
-				) : null}
+				{errorMessage || !searchResults ? (
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:29]: I wouldn't couple the missing search results with the error. I would expect either one of the other to be true at a given time. I would expect a message to render when there are no errors but also no search results because right now there is just a blank screen.
It's probably easier to check for `searchResults.length > 0`. and render "No results found" in that case.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx
```diff
@@ -71,12 +91,34 @@ export async function getServerSideProps(context) {
...
+			  ).data.map((value) => {
+					// restructure response to match expected article object structure
+					return value.attributes;
+			  })
+			: null;
+
+		// Do not send error unless search term is present so /search remains a valid path
+		if (!searchResults && searchTerm) {
+			errorMessage = true;
+		}
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:112]: I don't really understand this. It is possible to get 0 search results but there is no error.

CobyPear (changes requested):

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx
```diff
@@ -36,7 +68,6 @@ export async function getServerSideProps(context) {
 		res,
 		query: { alias },
```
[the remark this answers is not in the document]
Comment (mitchellmarkoff) [reply packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:69]: I originally tried destructuring like this but an error gets thrown if `alias` is undefined. `alias` is undefined anytime the user access  the root `/search` without a query.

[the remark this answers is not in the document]
Comment (CobyPear) [reply packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:69]: ah, in that case could you change 94 to the following
```js
const searchTerm = alias ? [alias] : null;
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx
Comment (CobyPear) [reply packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:112]: If I query 'test' I get 0 search results and no error, but also no message as to the search having been done or not.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx
Comment (CobyPear) [reply packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/[[...alias]].jsx:103]: nvm I think I misread the ternary, but I still think it might be better to check for `searchResults.length` instead of sending `searchResults || null` to the client.

mitchellmarkoff:
Per your comment [here](https://github.com/pantheon-systems/decoupled-kit-js/pull/619#discussion_r1162854360)

> This way I think we could go back to the single try/catch wrapping both calls, and still fail gracefully in case one or the other fetch calls fail.

This current implementation I sent back will fail gracefully if either of the fetch calls fail in a single try/catch however we may need to go back to two try/catch statements if we wish to differentiate between which one is failing. If `getDrupalSearchResults` fails, it throws an error rather than setting `searchResults` as undefined.

PR #623, opened 2023-04-12
DB-5241: tailwindless gatsby-wp templates

CobyPear:
## What changes were made?
- Add tailwindless gatsby templates
- Refine addWithDiff so that only files changed are written. This should prevent files left behind in case of an abort midway through this action.
- Add css modules for gatsby templates
- Refine runLint so that it uses the --fix flag

gatsby-wp templates, new templates added to CLI and some edited.

tested new tailwindless starter with the watch script. Also checked the wtih-tailwind option for regressions.
## Additional information


mitchellmarkoff (changes requested):
Overall this looks really great! No problems came up in testing however I think there are some styling differences in the pagination component that may be worth addressing. Kind of a nit but it appears the tailwindless Paginator loses the functionality of 
1. Highlighting the button of the current page selected.
2. Greying out either the forward or back buttons if the user is on the first or last page.

I think addressing at least bullet 1 above would be nice for the user to keep track of their current page but if that ask requires a lengthy or cumbersome implementation I am open to hearing thoughts against it.

Everything else is looking spot on, great work (:

CobyPear:
> Overall this looks really great! No problems came up in testing however I think there are some styling differences in the pagination component that may be worth addressing. Kind of a nit but it appears the tailwindless Paginator loses the functionality of
> 
> 1. Highlighting the button of the current page selected.
> 2. Greying out either the forward or back buttons if the user is on the first or last page.
> 
> I think addressing at least bullet 1 above would be nice for the user to keep track of their current page but if that ask requires a lengthy or cumbersome implementation I am open to hearing thoughts against it.
> 
> Everything else is looking spot on, great work (:

For the grayed out buttons, this doesn't appear to be working on the tailwind version of the gatsby-wp starter. I think the gatsby paginator component in general needs some refactoring that should make these buttons actually disabled when on page 1 or the last page. It should be easier to style them when they have the disabled attribute properly attached.

mitchellmarkoff (approved):
LGTM! 🕺

PR #636, opened 2023-04-18
DB-5716: Escape Characters in Search Queries

mitchellmarkoff:
## What changes were made?

- Change URL encoding to escape a larger set of characters.

`drupal-kit`

Searches with special characters such as '&' no longer break the `/search` route.

## Additional information


CobyPear (commented):
This seems to be working for cases like `&milk`, but single symbols like `&` are still returning all articles. Is this confirmed to be the intended behavior?

mitchellmarkoff:
> This seems to be working for cases like `&milk`, but single symbols like `&` are still returning all articles. Is this confirmed to be the intended behavior?

@CobyPear Ah yes I should have highlighted that, single symbols like `&` returning all articles was confirmed to be an acceptable behavior. Only needed to address cases like `&milk` or `milk & cookies`.

CobyPear (approved):

PR #639, opened 2023-04-18
[docs] Docs update npm init

CobyPear:
## What changes were made?
- Docs with `npm init` commands that use args have been updated to include the double dash that is required to forward args to the command
- Generated api reference - I meant to do this last sprint and we're a bit behind. Seems like an OK time to do this.
`web/docs`, project generator READMEs

## Additional information


mitchellmarkoff (approved):
LGTM 🕺

PR #640, opened 2023-04-19
DB-5758: Search Box Refinements

mitchellmarkoff:
## What changes were made?
- Updated the styling of the `<SeachInput />` component

The updated component was tested locally and seen to have correct styling and breakpoints.

## Additional information


CobyPear (changes requested):
The search box + button look good, but the nav bar seems to have had some regressions.

1. On smaller viewport widths, the emojis stack on top of the link text (see screenshot) It seems mostly fine on very small viewports.

Screen Shot 2023-04-19 at 3 49 59 PM

2. On wider viewports, the nav bar is not aligned with the rest of the page.

image

CobyPear (approved):
LGTM! 

One nit but no need to change now: Ideally we shouldn't need the overrides for the button's bg style. It looks like something is giving buttons with `type='submit'` a transparent bg. Kind of moot right now since we will be transitioning to css modules sooner or later.

PR #604, opened 2023-03-30
DB-5227: tailwindless next-drupal templates (part 2)

CobyPear:
## What changes were made?
- Add tailwindless next-drupal templates
- Add tailwindless next-drupal-umami-addon temlates
- Add subdirectorys in partils for better organization
- Fixed some style bugs in shared layout CSS
- Minor css module styles of search-addon header

## Where were the changes made?
cli > templates > partials, next-drupal, next-drupal-umami-addon, tailwind-shared tailwindless-next

## Additional information

Don't forget to [add a changeset](https://github.com/pantheon-systems/decoupled-kit-js#generating-a-changeset) if needed!

PR #647, opened 2023-04-24
DB-5773: CMS Type Option

mitchellmarkoff:
## What changes were made?

- A `--cmsType` flag can be passed to filter generators by their CMS compatibility
- A CMS Type prompt exists to take user input and filter suggested generators appropriately 
- This option is documented in cli help command

`cli`

This new option has been seen to work at sorting generators locally.

## Additional information

### DIFF
--- a/packages/create-pantheon-decoupled-kit/__tests__/index.test.ts
+++ b/packages/create-pantheon-decoupled-kit/__tests__/index.test.ts
@@ -4,10 +4,7 @@ import * as utils from '../src/utils/index';
 const { parseArgs, main } = bin;
 import chalk from 'chalk';
 import inquirer from 'inquirer';
-import {
-	decoupledKitGenerators,
-	sharedDecoupledKitGenerators,
-} from '../src/generators';
+import { decoupledKitGenerators } from '../src/generators';
 import versions from '../src/pkgVersions.json';
 
 import pkg from '../package.json' assert { type: 'json' };
@@ -65,11 +62,7 @@ describe('main()', () => {
 		// --help
 		process.argv = ['node', 'bin.js', '--help'];
 
-		await main(
-			parseArgs(),
-			decoupledKitGenerators,
-			sharedDecoupledKitGenerators,
-		);
+		await main(parseArgs(), decoupledKitGenerators);
 
 		expect(logSpy).toHaveBeenLastCalledWith(
 			utils.helpMenu(decoupledKitGenerators),
@@ -78,11 +71,7 @@ describe('main()', () => {
 		// -h
 		process.argv = ['node', 'bin.js', '-h'];
 
-		await main(
-			parseArgs(),
-			decoupledKitGenerators,
-			sharedDecoupledKitGenerators,
-		);
+		await main(parseArgs(), decoupledKitGenerators);
 
 		expect(logSpy).toHaveBeenLastCalledWith(
 			utils.helpMenu(decoupledKitGenerators),
@@ -95,22 +84,14 @@ describe('main()', () => {
 		// --version
 		process.argv = ['node', 'bin.js', '--version'];
 
-		await main(
-			parseArgs(),
-			decoupledKitGenerators,
-			sharedDecoupledKitGenerators,
-		);
+		await main(parseArgs(), decoupledKitGenerators);
 
 		expect(logSpy).toHaveBeenLastCalledWith(`v${pkg.version}`);
 
 		// -v
 		process.argv = ['node', 'bin.js', '-v'];
 
-		await main(
-			parseArgs(),
-			decoupledKitGenerators,
-			sharedDecoupledKitGenerators,
-		);
+		await main(parseArgs(), decoupledKitGenerators);
 
 		expect(logSpy).toHaveBeenLastCalledWith(`v${pkg.version}`);
 	});
@@ -127,7 +108,6 @@ describe('main()', () => {
 		await bin.main(
 			bin.parseArgs(process.argv.slice(2)),
 			decoupledKitGenerators,
-			sharedDecoupledKitGenerators,
 		);
 
 		expect(parseSpy).toBeCalled;
@@ -149,7 +129,6 @@ describe('main()', () => {
 				v: false,
 			},
 			decoupledKitGenerators,
-			sharedDecoupledKitGenerators,
 		);
 		expect(promptSpy).toHaveBeenCalledTimes(2);
 	});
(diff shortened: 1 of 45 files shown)


CobyPear (changes requested):
I should have linked this comment to the original ticket, because I think it informs the way this feature should be implemented: https://github.com/pantheon-systems/decoupled-kit-js/pull/604#discussion_r1157241140

So we should be refactoring any generators that include a cmsType in their `data` field. Some have `wp: true` or `drupal: true`. Instead, we should add a new field to the DecoupledKitGenerator interface so that it includes a `cmsType`. This type can be "wp", "drupal", or "any". Not the type `any`, but string literal "any". We could call the last one "either" but we might support more CMSes one day in the future.

This way, each generator can declare itself as being for wp, drupal, or either. So for the `main` logic, we will simply show all generators by default, and only filter by cmsType if that var exists.

mitchellmarkoff:
@CobyPear After trying to work with `cmsType` in our `hbs` statements I think it's best to stick with our use of the `wp` and `drupal` keys for the time being. I was able to expose the `cmsType` to handlebars but that, along with needing helper functions in handlebars to manage the variable, was adding a lot of bloat and complication that I think was unnecessary / not worth the effort since lots of this will get replaced soon.

packages/create-pantheon-decoupled-kit/src/types.ts
```diff
@@ -53,6 +53,10 @@ export interface DecoupledKitGenerator<
 	 * Any message(s) to be rendered after actions are successfully completed.
 	 */
 	nextSteps?: string[];
+	/**
+	 * Identifies a generators compatible CMS(s).
+	 */
+	cmsType: string;
```
CobyPear, line 59: This type can be more specific since we only have a specific set of cmsTypes
```suggestion
	cmsType: 'wp' | 'drupa' | 'any';
```

packages/create-pantheon-decoupled-kit/src/index.ts
```diff
@@ -85,11 +85,29 @@ export const main = async (
...
+
+		let suggestedGeneratorNames: string[] | null;
+
+		// filter generators if a cmsType was provided
+		if (cmsType) {
+			const suggestedGenerators = generators.filter((generator) => {
+				return generator.cmsType === cmsType || generator.cmsType === 'any';
+			});
+			suggestedGeneratorNames = suggestedGenerators.map(({ name }) => name);
+		}
```
CobyPear, line 101: This can be a bit more concise, and the name of the generator names could be more specific.
```suggestion
		const generatorsOfCmsType = generators
			.filter((generator) => {
				return isString(args.cmsType)
					? generator.cmsType === args.cmsType.toLowerCase() ||
							generator.cmsType === 'any'
					: true;
			})
			.map(({ name }) => name);
```

packages/create-pantheon-decoupled-kit/src/index.ts
```diff
@@ -85,11 +85,29 @@ export const main = async (
...
+
+		const whichGenerators: QuestionCollection<{
+			generators: string[];
+		}> = {
 			name: 'generators',
 			type: 'checkbox',
 			message: 'Which generator(s) would you like to run?',
-			choices: () => generatorNames,
+			choices: () =>
+				suggestedGeneratorNames ? suggestedGeneratorNames : generatorNames,
```
CobyPear, line 110: ... and this line becomes:
```suggestion
			choices: () =>
				generatorsOfCmsType.length > 0 ? generatorsOfCmsType : generatorNames,
```

packages/create-pantheon-decoupled-kit/src/index.ts
```diff
@@ -85,11 +85,29 @@ export const main = async (
 	// If no generators are found in positional params
 	// ask which generators should be run
 	if (!foundGenerators || !foundGenerators.length) {
-		const whichGenerators: QuestionCollection<{ generators: string[] }> = {
+		// check if a cmsType was provided
```
CobyPear, line 88: There should probably be some feedback in case the user chooses an invalid cms type like `wordpress` or `WordPress`, or even typos like `drpual`.  Check if the type is valid, and if not log a message that says something like `"invalid cmsType: ${cmsType}. Showing all generators"`.

web/docs/Frontend Starters/create-pantheon-decoupled-kit.md
```diff
@@ -108,6 +108,8 @@ npm init pantheon-decoupled-kit -- next-drupal next-drupal-umami-addon \
   # prevent the package manager from running
   # `install` after the project is created
   --noInstall
+  # CMS you are developing for. This flag is used to filter suggested generators by CMS compatibility. As of now, `wp` or `drupal` are valid arguments
+  --cmsType drupal
```
CobyPear, line 112: I'm not sure the flag is necessary on this doc because the generators are already selected.

CobyPear (changes requested):
Mostly minor feedback on some code style, and I think we could have a message logged in case the cms type is invalid similar to when you choose an invalid generator via positinal params.

There should probably be one or two new tests cases supporting this feature as well.

packages/create-pantheon-decoupled-kit/src/index.ts
mitchellmarkoff: Used your suggestion as a base and added a few cases to handle error messaging and avoid returning only generators with type `any` if an incorrect `cmsType` was provided

packages/create-pantheon-decoupled-kit/__tests__/index.test.ts
```diff
@@ -269,4 +271,53 @@ To see this list at any time, use the --help command.`);
...
+			choices: [
+				'next-wp',
+				'gatsby-wp',
+				'next-drupal',
+				'next-drupal-umami-addon',
+				'next-wp-acf-addon',
+				'gatsby-wp-acf-addon',
+				'tailwindcss-addon',
+				'next-drupal-search-api-addon',
+			],
```
CobyPear, line 317: These tests look good! Just one change – make this list dynamic, same with the 'filtered' list. This way we won't have to update the tests when we add new generators

packages/create-pantheon-decoupled-kit/src/index.ts
```diff
@@ -85,11 +85,34 @@ export const main = async (
 	// If no generators are found in positional params
 	// ask which generators should be run
 	if (!foundGenerators || !foundGenerators.length) {
-		const whichGenerators: QuestionCollection<{ generators: string[] }> = {
+		let generatorsOfCmsType: string[] = [];
+		const cmsType = isString(args.cmsType)
+			? args.cmsType.toLocaleLowerCase()
```
CobyPear, line 90:
```suggestion
			? args.cmsType.toLowerCase()
```
Shouldn't need the Locale distinction here

packages/create-pantheon-decoupled-kit/src/index.ts
```diff
@@ -85,11 +85,34 @@ export const main = async (
...
+				console.log(
+					chalk.yellow(`Invalid cmsType: ${cmsType}. Showing all generators.`),
+				);
+			} else {
+				generatorsOfCmsType = generators
+					.filter((generator) => {
+						return generator.cmsType === cmsType || generator.cmsType === 'any'
+							? true
+							: false;
+					})
```
CobyPear, line 104:
```suggestion
					.filter((generator) => generator.cmsType === cmsType || generator.cmsType === 'any'
					})
```

Since you split out the other logic for the error handling, you can use an implicit return here and get rid of the ternary. idc as much about the implicit return, but the ternary should definitely go here :)

CobyPear (changes requested):
A few little things but it's looking very close

PR #661, opened 2023-05-03
DB-5692: Refactor `cms-kit` bundle

CobyPear:
## What changes were made?
I opened up a new branch, `refactor-bundles` to pool some of this work for smaller PRs

This first one refactor's the `cms-kit` bundle. The rest of the packages will follow a similar pattern
- Remove default exports - named exports only from now on unless default export is absolutely required
- Restructure the `src` folder a bit - types (if any) should go in `types.ts` or in the file they are needed if they aren't shared. This is instead of `types/index.ts` which makes extra dirs in the final bundle that we don't really need.
- Renamed `utils` to `lib` to keep our exported dirs consistent across packages
- Remove multiple barrel files. Only one barrel file is present at the root of ./src
- Use `tsup` for the bundling instead of vite. `tsup` is using the same stuff under the hood, but it seems to handle a few things a bit better and I am having less trouble with it in general compared to vite (for building).
- Use a different tsconfig for building vs developing - this is so we can keep our tests included in the tsconfig files so they are properly typechecked.
- package.json - `type: module` is now in use. Exports field has been redone to satisfy `npx publint`

`cms-kit`

## Additional information

This is the first PR in this series.


mitchellmarkoff (approved):
LGTM 🕺

pnpm-lock.yaml
```diff
@@ -2840,7 +2837,7 @@ packages:
     peerDependencies:
       react: '*'
     dependencies:
-      '@types/react': 17.0.40
```
CobyPear, line 2843: I will look into why this is being reverted on my machine
cc @mitchellmarkoff

CobyPear: I downgraded the `@types/react` in `web` to v17 which should be consistent with what is there now.

PR #663, opened 2023-05-04
DB-5950: `convertCssModules` Action

mitchellmarkoff:
## What changes were made?
- A `convertCssModule` action has been created to handle the process of converting CSS modules to tailwind. This action makes use of the [css-modules-to-tailwind](https://www.npmjs.com/package/css-modules-to-tailwind), [postcss-custom-properties](https://www.npmjs.com/package/postcss-custom-properties), [@csstools/postcss-global-data](https://www.npmjs.com/package/@csstools/postcss-global-data), and [postcss-cli](https://www.npmjs.com/package/postcss-cli) packages.
- CSS Module conversion is only supported through use of the `next-drupal-search-api-addon` 
- A bug was fixed in the `<SearchInput />` component. If a user viewed a Page via the /[...alias] route, the Page title would show up in the search bar. Changing the search page naming to a different catch all route fixed this issue.

`cli < templates < next-drupal, next-drupal-search-api-addon, partials, tailwind-shared`
`cli < generators < next-drupal, next-drupal-search-api-addon`
`cli < actions`
`cli < tests`

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/partials/next-drupal/nextDrupalPkgJson.hbs
```diff
@@ -0,0 +1,48 @@
...
+		"start": "node .next/standalone/server.js",
+		"build:mono": "next build",
+		"start:mono": "next start",
+		"lint": "next lint",
+		"test": "vitest run",
+		"update-snapshots": "vitest run --update --silent",
+        {{#if tailwindcss}}
+        {{#if convertCssModules}}
+        "translate-vars": "postcss components/*.css pages/**/*.css --replace",
+		"convert-to-tailwind": "css-modules-to-tailwind pages/**/*.jsx  components/*.jsx --force",
```
CobyPear, line 15: I'm not sure about this approach. Ideally, we would not need to leave extra scripts in the output. I would expect the action to run these commands on the output instead. That way we could keep all of the dependencies in the CLI instead of adding one-time use packages to the user's output.

I think the switch is not so bad – add the dependencies to the CLI and call `npx postcss components/*.css pages/**/*.css --replace` and `npx css-modules-to-tailwind pages/**/*.jsx components/*.jsx --force` using `execSync` like we're doing for these scripts.

packages/create-pantheon-decoupled-kit/src/generators/next-drupal-search-api-addon.generator.ts
```diff
@@ -21,12 +27,14 @@ export const nextDrupalSearchApiAddon: DecoupledKitGenerator<
 	addon: true,
 	data: {
 		search: true,
+		convertCssModules: true,
```
CobyPear, line 30: We should be able to nix this and use the `tailwindcss`/`noTailwindcss` data prop coming from the tailwind prompt, because it will be css modules by default and tailwind (by way of conversion) otherwise.

packages/create-pantheon-decoupled-kit/src/templates/tailwind-shared/postcss.config.cjs.hbs
```diff
@@ -0,0 +1,16 @@
...
+		tailwindcss:{},
+		autoprefixer:{},
+		{{#if convertCssModules}}
+		'@csstools/postcss-global-data':
+		{ 
+		files: [
+			'./styles/globals.css',
+		],
+		},
+		'postcss-custom-properties':{ preserve: false },
```
CobyPear, line 12: Can these options be supplied via the postcss CLI when it is called instead? Thinking about how to keep all the deps for this out of the final output.
If necessary you may need to write some TS that uses these packages directly to do the transform and rewrite it to the file system, like we are doing in `addWithDiff`.

CobyPear (changes requested):
Good start on this one. I think the implementation shouldn't leave extra packages/config in the output if possible. I was thinking that we would use these packages in the CLI itself to do the transformation, or calling their `bin`'s through `npx` instead of using scripts in the `package.json`s.
LMK if you need more detail on either approach. I would start with `npx` since it's almost there already.

.changeset/strong-knives-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-pantheon-decoupled-kit": patch
```
CobyPear, line 2: I think this is worthy of a minor change.

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -0,0 +1,71 @@
...
+	if (data?.noInstall || !data?.tailwindcss)
+		return 'skipping CSS module conversion';
+	if (!data.outDir || !isString(data?.outDir))
+		throw new Error('outDir is not valid');
+	data.silent || console.log(chalk.green('Converting Modules...'));
+
+	try {
+		const allFiles: string[] = [];
+		const componentFiles = await getAllFiles(
+			`${data.outDir}/components`,
```
CobyPear, line 38: I prefer `path.join(data.outDir, 'components')` here. The template literals for dirs are better for strings, not necessarily paths to be found.

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -0,0 +1,71 @@
...
+			data.outDir,
+		);
+		const pageFiles = await getAllFiles(`${data.outDir}/pages`, data.outDir);
+		allFiles.push(...componentFiles);
+		allFiles.push(...pageFiles);
+		for (const file of allFiles) {
```
CobyPear, line 44:
```suggestion
		for (const file of klaw(outDir)) {
```
If you make this change here, I don't think the previous lines gathering up `allFiles` are necessary. We already know all of the files are going to be in `outDir`, so we shouldn't need extra logic to determine what files to read.

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -0,0 +1,71 @@
...
+			if (file.includes('.module.css')) {
+				const contents = fs.readFileSync(`${data.outDir}${file}`).toString();
```
CobyPear, line 46:
```suggestion
				const contents = fs.readFileSync(`${data.outDir}${file}`, 'utf-8');
```
You shouldn't ever need to `.toString()` a `readFile` method. You can choose the encoding type as an option in the second arg. This way it returns a `utf-8` encoded string instead of a plain buffer.

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return templates;
+};
+
+export const convertCSSModules: Action = async ({ data }) => {
+	if (data?.noInstall || !data?.tailwindcss)
+		return 'skipping CSS module conversion';
+	if (!data.outDir || !isString(data?.outDir))
+		throw new Error('outDir is not valid');
+	data.silent || console.log(chalk.green('Converting Modules...'));
```
CobyPear, line 33:
```suggestion
	data.silent || console.log(chalk.green('Converting CSS modules to tailwindcss classes...'));
```
"Modules" by itself is a bit vague.

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -0,0 +1,71 @@
...
+
+	// loop through each dir
+	for await (const file of klaw(templateDir)) {
+		if (file.stats.isDirectory()) continue;
+		const templateName = file.path.split(outDir)[1];
+
+		templates.push(templateName);
+	}
+	return templates;
+};
```
CobyPear, line 26: See comment on line 44 as I'm not sure this extra helper is needed. If it is needed:
I don't mind this logic being broken out into a helper function, but it can be written in a more generic way.
1. if it's truly a generic getAllFiles helper, why does it need two args? It should probably take 1 arg (a path to a directory) and return all files in that dir. 
2. In this case, we should be gathering already rendered _files_ not _templates_. So there should be no need to split the template name from the file path like we do in `dedupeTemplates`.
3. I would also move this to `utils` if it's meant to be a general helper utility, otherwise you can ditch the `export` and just use it locally in this file.

packages/create-pantheon-decoupled-kit/__tests__/convertCSSModules.test.ts
CobyPear, line 1: The test coverage shows a bit of a gap on the actual processing part. I think a spy on `postcss` to make sure it was called correctly would be a good test to have..

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -0,0 +1,71 @@
...
+		const allFiles: string[] = [];
+		const componentFiles = await getAllFiles(
+			`${data.outDir}/components`,
+			data.outDir,
+		);
+		const pageFiles = await getAllFiles(`${data.outDir}/pages`, data.outDir);
+		allFiles.push(...componentFiles);
+		allFiles.push(...pageFiles);
+		for (const file of allFiles) {
+			if (file.includes('.module.css')) {
```
CobyPear, line 45: Small nit: we might want to be more specific here and use `endsWith`.
```suggestion
			if (file.endsWith('.module.css')) {
```

CobyPear (changes requested):
This is really slick! Besides for the inline comments, on my test output it seems like the `searchInput.module.css` and the `search-input.jsx` did not get the tailwindcss treatment:


<details>
  <summary>Here is what was left in `search-input.jsx`</summary>

```jsx
<form onSubmit={onSearch}>
			<div
				className={`${styles.container} items-stretch flex pt-9 relative w-full`}
			>
				<label htmlFor="search">
					<input
						className={`${styles.searchInput} bg-transparent rounded-l border-solid border text-base font-normal h-full px-3 duration-200 ease-in-out`}
						type="search"
						value={searchQuery || ''}
						onChange={(event) => setSearchQuery(event.target.value)}
						placeholder="Search"
						id="search"
						aria-label="Search Bar"
					/>
				</label>

				<button
					className={`${styles.searchBtn} items-center rounded-r text-white flex py-2.5 px-6 relative`}
					type="submit"
					id="submit-btn"
					aria-label="Submit Search"
				>
					<svg
						xmlns="http://www.w3.org/2000/svg"
						viewBox="0 0 20 20"
						fill="currentColor"
						className={`${styles.icon} h-5 w-5`}
```
</details>

<details>
  <summary>And here is the what was in the searchInput.module.css</summary>

  ```css
.searchInput { 
    transition-property: background-color, border-color, color, fill, stroke, opacity, box-shadow, transform; 
    outline: 0; 
}

.searchBtn { 
    background-color: rgb(37 99 235) !important; 
    box-shadow: 0 0.25rem 0.375rem -(1px) rgba(0, 0, 0, 0.1), 0 0.125rem 0.25rem -(1px) rgba(0, 0, 0, 0.06); 
}

.searchBtn:hover {
    @apply bg-blue-700;
    box-shadow: 0 0.625rem 1rem -(0.125rem) rgba(0, 0, 0, 0.1), 0 0.25rem 0.375rem -(0.125rem) rgba(0, 0, 0, 0.05);
}

  ```
</details>

packages/create-pantheon-decoupled-kit/__tests__/convertCSSModules.test.ts
mitchellmarkoff: Was not able to get this working in the end so I left it out of these revisions.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/components/searchInput.module.css
```diff
@@ -24,23 +24,27 @@
...
     align-items: center; 
     border-top-right-radius: var(--1);
     border-bottom-right-radius: var(--1); 
-    box-shadow: var(--0) var(--1) var(--1_5) -(var(--px)) rgba(0, 0, 0, 0.1), var(--0) var(--0_5) var(--1) -(var(--px)) rgba(0, 0, 0, 0.06); 
+    box-shadow: var(--0) var(--1) var(--1_5) calc(var(--px) * -1) rgba(0, 0, 0, 0.1), var(--0) var(--0_5) var(--1) calc(var(--px) * -1) rgba(0, 0, 0, 0.06); 
 }
 
 .searchBtn:hover {
-    box-shadow: var(--0) var(--2_5) var(--4) -(var(--0_5)) rgba(0, 0, 0, 0.1), var(--0) var(--1) var(--1_5) -(var(--0_5)) rgba(0, 0, 0, 0.05); 
+    box-shadow: var(--0) var(--2_5) var(--4) calc(var(--0_5) * -1) rgba(0, 0, 0, 0.1), var(--0) var(--1) var(--1_5) calc(var(--0_5) * -1) rgba(0, 0, 0, 0.05);
```
CobyPear, line 47: the `calc(val -1)` is harder to read then what you had before imo.

CobyPear (changes requested):
One inline comment, but also can you change the `.container > ul h2` selector in `searchPage.module.css` to its own class like `.listTitle` and use that class in `[[...searchAlias]].jsx`? We can start replacing some of those nested selectors with classes to help out the css conversion.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/components/searchInput.module.css
mitchellmarkoff: @CobyPear I don't know how I did not notice this earlier but actually what I had there before was not working `-(var(--3))` was not registering as valid and was breaking the style. This way seems to work fine 👍

CobyPear: ooh ok gotchya, works for me.

CobyPear (approved):
🚀

PR #665, opened 2023-05-05
DB-6013: Update `setSurrogateKeyHeader` and `drupal-kit` `defaultFetch`

CobyPear:
## What changes were made?
- update `setSurrogateKeyHeader` to remove the first two keys if there are more than 2. I'm not positive this logic is correct for all cases and it's hard to debug the hashed values because they are hashed. I was going off of the assumption that the test input wouldn't change, but maybe the `resourceKeys` in the `cms-kit` test is no longer valid?
- Update `defaultFetch` to get and set the `Surrogate-Key` header instead of `Surrogate-Key-Raw`
Tested locally with a generated starter inside the monorepo.
## Additional information

Fastly sets `Surrogate-Key` not `Surrogate-Key-Raw`. I could see the desire to do the hashing ourselves if debugging becomes an issue.

### DIFF
--- a/packages/cms-kit/src/utils/setSurrogateKeyHeader.ts
+++ b/packages/cms-kit/src/utils/setSurrogateKeyHeader.ts
@@ -18,12 +18,19 @@ const setSurrogateKeyHeader = (
 	keys: string | null,
 	res: ServerResponse,
 ): string | void => {
+	const removeFirstTwo = (str: string) => {
+		const arr = str.split(' ');
+		const result = arr.length > 2 ? arr.slice(2) : arr;
+		return result;
+	};
+
 	if (hasHeaders(res)) {
-		const surrogateKeyHeader = res.getHeader('Surrogate-Key-Raw');
+		const surrogateKeyHeader = res.getHeader('Surrogate-Key');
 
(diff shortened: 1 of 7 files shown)


packages/cms-kit/src/utils/setSurrogateKeyHeader.ts
```diff
@@ -18,22 +18,30 @@ const setSurrogateKeyHeader = (
...
 	res: ServerResponse,
 ): string | void => {
+	const removeFirstTwo = (str: string) => {
+		const arr = str.split(' ');
+		const result = arr.length > 2 ? arr.slice(2) : arr;
+		return result;
+	};
+
 	if (hasHeaders(res)) {
 		const surrogateKeyHeader = res.getHeader('Surrogate-Key-Raw');
```
CobyPear, line 28: I'm actually not sure if this needs to be changed to `Surrogate-Key` cc @backlineint

CobyPear: Nvm, I think it makes sense to change it but I can change it back if needed.

backlineint (approved):

PR #680, opened 2023-05-15
DB-6036: [gatsby-wp] Use css modules for default stying

CobyPear:
## What changes were made?
- Add react/jsx-runtime rule to react eslint config
- Remove tailwind classes from components and templates in gatsby-wp templates
- Remove tailwindless gatsby templates
- Use `convertCSSModules` for `gatsby-wp` the tailwindcss version.
- Refactor `convertCSSModules` action to work with gatsby and without install
- Refinement of error handling
 
`eslint-config`, `cli`, `gatsby-wp` tempaltes 

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/src/components/footer.module.css
```diff
@@ -24,27 +20,31 @@
...
+
 .footerCopy {
 	margin: var(--4) auto;
 	padding: var(--2);
 	color: var(--white);
 }
 
-.footerCopy > a {
+.footerCopyLink {
 	text-decoration: underline;
```
mitchellmarkoff, line 48: `text-decoration` -> `text-decoration-line`

packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/src/components/footer.module.css
```diff
@@ -24,27 +20,31 @@
...
-.footerNav > ul li {
-	list-style: disc;
-	color: var(--blue-300);
+.footerUl {
+	margin-top: var(--4);
 }
-
 .navItem {
 	margin-left: var(--3);
+	list-style: disc;
```
mitchellmarkoff, line 28: `list-style` -> `list-style-type`

packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/src/components/footer.module.css
```diff
@@ -24,27 +20,31 @@
 	margin: var(--x-auto);
```
mitchellmarkoff, line 20: since the value of `--x-auto` is `0 auto`, it will not get converted, tailwind would understand plain `auto`

mitchellmarkoff (changes requested):
Great work here! The only thing I really see to point out is the updating of CSS values to match those in tailwind so more will get converted. I left comments with examples of some attributes that can be edited. If you add/edit any of our own CSS vars, they will need to be edited in both `/create-pantheon-decoupled-kit/src/templates/gatsby-wp/src/css/style.css` AND `/create-pantheon-decoupled-kit/src/templates/styles/variables.css`

_CSS attributes and their values must match a tailwind default to get converted. If a value is not converted, check that it exists in tailwind._

One other thing is that we're now getting these warnings in the console when running dev mode. Not sure if there is anything we can necessarily do about these but wanted them documented here.
 
Screenshot 2023-05-15 at 11 25 55 AM

CobyPear:
> One other thing is that we're now getting these warnings in the console when running dev mode. Not sure if there is anything we can necessarily do about these but wanted them documented here.

Oh shoot I forgot about this. Gatsby complains about your CSS modules if components are not imported in the perfect order in each template 😭. I'll try to clean it up. I really wanted to add the vscode auto order imports option to the repo but I guess it won't work because of this.

CobyPear:
> > One other thing is that we're now getting these warnings in the console when running dev mode. Not sure if there is anything we can necessarily do about these but wanted them documented here.
> 
> Oh shoot I forgot about this. Gatsby complains about your CSS modules if components are not imported in the perfect order in each template 😭. I'll try to clean it up. I really wanted to add the vscode auto order imports option to the repo but I guess it won't work because of this.

Some of these are due to the `css-modules-to-tailwind` package leaving some classnames around even though they are completely removed from the css module file. I'll try to clean up any that I can but some of these might be unavoidable. 


I'm a bit worried about leaving so much stuff behind with this conversion process, but outside of using a different package for the conversion there's not much to do about it.

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -39,18 +38,22 @@ export const convertCSSModules: Action = async ({ data }) => {
 			});
 			fs.writeFileSync(file, result);
 		}
+		const dirs = data.gatsby
+			? './src/**/*.jsx'
+			: './pages/**/*.jsx ./components/*.jsx';
 		execSync(
-			`npx css-modules-to-tailwind pages/**/*.jsx components/*.jsx --force`,
+			`npx --prefer-online --yes css-modules-to-tailwind ${dirs} --force`,
```
CobyPear, line 45: @mitchellmarkoff This should make sure `npx` always fetches from the registry. Eventually we can abstract out a `runNpx` or `runPackageCommand` type helpers to make the `execSync` stuff a bit nicer

packages/create-pantheon-decoupled-kit/src/actions/convertCSSModules.ts
```diff
@@ -1,17 +1,17 @@
...
 import { globSync } from 'glob';
 import path from 'path';
+import postcss from 'postcss';
+import customProperties from 'postcss-custom-properties';
+import { rootDir } from '..';
+import { Action, isString } from '../types';
 
 export const convertCSSModules: Action = async ({ data }) => {
-	if (data?.noInstall || !data?.tailwindcss)
+	if (data.noInstall || !data?.tailwindcss)
```
mitchellmarkoff, line 14:
```suggestion
	if (data?.noInstall || !data?.tailwindcss)
```

mitchellmarkoff (changes requested):
Really close, I am just seeing one issue with the footer styling. It is displaying like this in both tailwind and tailwindless generated projects
Screenshot 2023-05-15 at 2 33 07 PM

mitchellmarkoff:
> I'm a bit worried about leaving so much stuff behind with this conversion process, but outside of using a different package for the conversion there's not much to do about it.

Yeah it is a bit of a bummer we're left with not much to do about it. I'm cool rolling with it for now 👍

mitchellmarkoff (approved):
LGTM! Nice work 🕺 🦈

PR #684, opened 2023-05-16
DB-6039: [gatsby-wp-acf-addon] Use css modules by default

CobyPear:
## What changes were made?
- Specify eslint ignore files in cli .eslintrc
- Add css modules for gatsby-wp-acf-addon
- Make page & post widths consistent

`cli`, `gatsby-wp-acf-addon` templates

Locally w/watch script
## Additional information
Added the `.jsx` files to the ignorePatterns in `.eslintrc` of the cli to avoid errors on those files. They are still able to be formatted by prettier. The gatsby acf addon however doesn't have the gatsby `.prettierrc` so if you format it it will add semicolons etc. In case you need to format that one file you can temporarly add the `.prettierrc` to the addon dir.

### DIFF
--- a/.changeset/strange-shrimps-learn.md
+++ b/.changeset/strange-shrimps-learn.md
@@ -1,0 +1,5 @@
+---
+'create-pantheon-decoupled-kit': patch
+---
+
+[gatsby-wp-acf-addon] Use css modules by default, retain tailwindcss option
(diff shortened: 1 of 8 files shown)


mitchellmarkoff (approved):
LGTM 🕺, ticket just needs manual test steps.

PR #710, opened 2023-05-25
DB-5244: Add tagged template support in addWithDiff + tests

CobyPear:
## What changes were made?
- Add support for tagged template literal templates
- Add helpers for the tagged template literal templates
  - These are the same as the handlebars helpers, but the types are slightly different so I figured it was worth separating them. Especially since we will deprecate and remove the handlebars stuff eventually.
- Add tests for the new template format

`cli` `addWithDiff`

tests!
## Additional information
Part 1 of DB-5244 in effort of keeping PRs more reviewable.

### DIFF
--- a/packages/create-pantheon-decoupled-kit/src/utils/taggedTemplateHelpers.ts
+++ b/packages/create-pantheon-decoupled-kit/src/utils/taggedTemplateHelpers.ts
@@ -1,0 +1,39 @@
+export const taggedTemplateHelpers = {
+	/**
+	 * @param condition - a single key of an {@link Input}
+	 * @param value - the value to render if the condition is true
+	 * @returns the value if it a string, or an empty string
+	 */
+	if: (condition: unknown, value: string) => (condition ? value : ''),
+	/**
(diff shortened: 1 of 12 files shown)


packages/create-pantheon-decoupled-kit/src/utils/taggedTemplateHelpers.ts
```diff
@@ -0,0 +1,39 @@
+export const taggedTemplateHelpers = {
+	/**
+	 * @param condition - a single key of an {@link Input}
+	 * @param value - the value to render if the condition is true
+	 * @returns the value if it a string, or an empty string
+	 */
+	if: (condition: unknown, value: string) => (condition ? value : ''),
```
CobyPear, line 7: This one is new and I expect we might have more helpers similar to this one to prevent things like 

```ts
${data.tailwindcss ? 'x' : ''}
``` 
It becomes 
```ts
${utils.if(data.tailwindcss, 'x')}
```
Open to different names.

mitchellmarkoff (approved):
LGTM 🕺 🦖

PR #711, opened 2023-05-26
DB-5870: Run Snapshot Tests With Git Hooks

mitchellmarkoff:
## What changes were made?
- When making a commit to the monorepo, starter kit snapshot tests will run.
- Large diff is a result of need to update snapshot tests
- Since the CSS modules -> tailwind package renders our styles in `${}` syntax, they are rendered as undefined in the snapshots. If we think generating starters with CSS modules instead of tailwind is preferred to avoid some of those missing values that would get us the full module name in the snapshots.

- `.husky/pre-commit`

All starter kit tests observed to run and pass locally when committing to the monorepo.

## Additional information


mitchellmarkoff:
Closing until a new development path for image based snapshot testing has been chosen.

PR #733, opened 2023-06-12
DRAFT: Update typedocs for latest codebase

backlineint:
## What changes were made?

* Committed updates to pnpm.lock
* Generated docs and corrected gh links.

Docs.

## Additional information

Still a draft as I'm trying to figure out if we need to do anything to address the failures shown here: https://app.warp.dev/block/pxCOtaIAMdNdUD8Ai1ePML


pnpm-lock.yaml
CobyPear, line 1: These changes seems strange to me, but it is likely no big deal. I suppose it could be the result of upstream dependencies changing.

backlineint: They were resulting in the install failing in CI as well, so I re-rolled without them.

CobyPear (approved):

PR #738, opened 2023-06-13
DB-6178: App Dir Header Implementation

mitchellmarkoff:
## What changes were made?
- Brought the `Header` component from `nextjs-kit` into the `next-drupal` app directory based starter.

`cli < templates < next-drupal-app` 

A new `next-drupal-app` was generated through the cli and observed to have the `Header` component in it's root layout.
## Additional information

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -12,5 +12,7 @@
     "@pantheon-systems/wordpress-kit": "3.0.0",
     "web": "1.0.0"
   },
-  "changesets": []
+  "changesets": [
+    "soft-tools-rescue"
+  ]
 }
--- a/packages/create-pantheon-decoupled-kit/CHANGELOG.md
+++ b/packages/create-pantheon-decoupled-kit/CHANGELOG.md
@@ -1,5 +1,11 @@
 # create-pantheon-decoupled-kit
 
+## 0.9.1-canary.0
+
+### Patch Changes
+
+- 2e9032db: [cli] Add aliases for wordpress and drupal CMS types.
+
 ## 0.9.0
 
 ### Minor Changes
--- a/packages/create-pantheon-decoupled-kit/package.json
+++ b/packages/create-pantheon-decoupled-kit/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "create-pantheon-decoupled-kit",
-	"version": "0.9.0",
+	"version": "0.9.1-canary.0",
 	"description": "Pantheon Decoupled Kit CLI",
 	"license": "GPL-3.0-or-later",
 	"homepage": "https://github.com/pantheon-systems/decoupled-kit-js#readme",


CobyPear:
@mitchellmarkoff @backlineint Quick question before diving into a full review: Is there a reason this is targeting `feat/app-directory-generators` and not the `feat/typescript-generators` branch?

mitchellmarkoff:
> @mitchellmarkoff @backlineint Quick question before diving into a full review: Is there a reason this is targeting `feat/app-directory-generators` and not the `feat/typescript-generators` branch?

I broke it off into it's own branch because in the beginning I wasn't sure if we would be fully removing the OG `next-drupal-starter` at the start of this app dir work. Ended up going the route of keeping it around though so I think it may be fine to have everything under one feature branch if we prefer. @backlineint any thoughts?

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/layout.tsx
```diff
@@ -1,16 +1,27 @@
-import { Inter } from 'next/font/google';
-import './globals.css';
+'use client';
```
CobyPear, line 1: I'm still wrapping my head around React Server Components, but I'm not sure we want this to be a client component.

From the docs (https://nextjs.org/docs/getting-started/react-essentials#client-components):

> Once "use client" is defined in a file, all other modules imported into it, including child components, are considered part of the client bundle.

So to my understanding, the `use client` directive on the layout would make the entire app client side rendered since this component is at the top of the tree.
We want layout to be a server component since it will include fetching for the footer menu at some point soon.

CobyPear (changes requested):
Only one comment/concern and it could just be my inexperience with RSC.
Also as we discussed, I would like to se a separate generator for the `next-drupal-app` templates if we are going to keep around the old templates while we work on this.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/layout.tsx
mitchellmarkoff: Oh yeah I was trying to avoid making this a client component but it was required with the use of the `nextjs-kit` header. Let me look into work arounds, thanks for the docs ref!

CobyPear: > it was required with the use of the nextjs-kit header.

Hmm that is a red flag to me. I dug into this and it looks like the nextjs-kit bundle is still a bit off, but when testing the correct bundle locally I kept running into this issue: https://github.com/vercel/next.js/issues/46267

The reason next is complaining is because the bundle is not correct, the file with the component also has hooks which are client only. So the error makes sense in that case, but when we split everything out it should be able to use only the Header component (which should work on the server).
 
I can open a draft PR to your branch with the bundle fix which we should do eventually, but it is not having much effect right now.

backlineint:
> I broke it off into it's own branch because in the beginning I wasn't sure if we would be fully removing the OG `next-drupal-starter` at the start of this app dir work. Ended up going the route of keeping it around though so I think it may be fine to have everything under one feature branch if we prefer. @backlineint any thoughts?

I was under the assumption that these were two branches because the non-app routing TypeScript work could be released first. We'd merge the TS work into the app routing branch on a regular basis. But if releasing the Gatsby TS work first isn't practical, we can combine these.

Let's check in on this in the standup today.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/layout.tsx
backlineint: Also don't have enough legit experience with RSC yet, but conceptually I actually can't think of anything that would need to be a client component in this new architecture. Am I missing something?

CobyPear: > I actually can't think of anything that would need to be a client component in this new architecture

Anything with a hook will need to be a client component. We are really only using hooks (`useRouter`, `useEffect`) for the "go back" link and for the paginator component. I think it's fine if those are client components and we can rearchitect them if needed. Unfortuently, next.js doesn't seem to understand our bundle and when it tries to import a component, it does so from the `index.mjs` bundle which includes ALL the components, so next thinks it's a client component because `useEffect` is imported to that file, even though it is not used in the `Header` component..

mitchellmarkoff: Agh that is unfortunate, thanks for digging into this @CobyPear. I have a few questions about this to bring up during standup.

backlineint: > Anything with a hook will need to be a client component.

Good callout. As I understand it, it isn't all hooks that can't be run on the server, but the rules are broad enough that it ends up being most of them. Based on the reliance on hooks I think this is going to break most people's brains when they migrate to server components.

> We are really only using hooks (useRouter, useEffect) for the "go back" link and for the paginator component.

With the app directory, I've been assuming that it would be possible to implement pagination server side.

And given this restriction, I wonder about the value of the 'go back' link. Would we just be better off with more traditional breadcrumbs that don't rely on routing?

CobyPear (approved):
LGTM! I am wondering if we should cherry-pick the `nextjs-kit` package changes to `canary` so that when we are testing this we don't have to symlink `nextjs-kit` and we can use the published canary instead.

CobyPear:
@mitchellmarkoff I think we should wait to merge this until we get the `nextjs-kit` into canary, then merge canary into this branch to avoid a potential conflict down the line. Does that make sense?

.changeset/hot-oranges-notice.md
CobyPear, line 1: @mitchellmarkoff I think I named the changeset incorrectly on the other branch and this one is showing up again, might need to manually delete it.

.changeset/dirty-garlics-switch.md
CobyPear, line 1: @mitchellmarkoff similarly this one might need a new name, my bad!

PR #749, opened 2023-06-21
DB-6357: Upgrade gatsby-wp templates to gastby v5

CobyPear:
## What changes were made?
- Updates to `gatsby-wp` templates for compatibility with gatsby 5
- Removes some unused/optional plugins
- Update gatsby dependencies
- Remove some unused dependencies
- Fix some css regressions and modules->tw issues

gatsby-wp templates

## Additional information
My hope was that the ESM support in `gatsby@5.3` would allow the `convertTStoJS` action to work without any extra codemods, however that is not the case.
The issue is that, after being converted to JS, the `import { something } from './somewhere'` syntax is not valid in ESM context, and the import path requires the `.mjs` extension, and the js files themselves will need the `.mjs` file to work as ESM. Still looking into potential solutions for that issue, but in the meantime this should be all that is required for the gatsby 5 upgrade.


mitchellmarkoff (changes requested):
All looks great but I am getting this one warning in both the tailwind and tailwindless versions of the project. Looking at the git history this class `componentContainer` was in earlier but it's definition got deleted at some point. Once that is added back to `./paginator.module.css` everything should be good to go 💯 
Screenshot 2023-06-22 at 4 14 25 PM

mitchellmarkoff (approved):
LGTM 🕺

PR #756, opened 2023-06-22
DB-6179 App Directory Root Layout with `<Footer />`

mitchellmarkoff:
## What changes were made?
- Brought the `Footer` component from `nextjs-kit` into the `next-drupal` app directory based starter.
- SSR data fetching introduced into the Root Layout

- A new `next-drupal-app` was generated through the cli and observed to have the `Footer` component in it's root layout.
## Additional information


packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/lib/stores.ts
```diff
@@ -0,0 +1,59 @@
...
+// getStaticProps and getServerSideProps
+/**
+ * @type {DrupalState[]}
+ * @remarks These stores have auth enabled. Authenticated and anonymous request control takes place on the level of the request.
+ * @see {@link https://project.pages.drupalcode.org/drupal_state/en/getting-objects/} for more information.
+ *
+ */
+export const globalDrupalStateStores = makeLocaleStores({
+	locales: process.env.locales as unknown as string[],
+	debug: process.env.DEBUG_MODE as unknown as boolean || false,
```
mitchellmarkoff, line 36: When converting this file to TS I landed this typing `process.env.DEBUG_MODE as unknown as boolean`. I feel like it smells a bit weird though, lmk if you have any thoughts.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/lib/stores.ts
```diff
@@ -0,0 +1,59 @@
...
+					debug: debug,
+					clientId: process.env.CLIENT_ID,
+					clientSecret: process.env.CLIENT_SECRET,
+				}),
+		  ];
+
+// exporting so we can reuse in
+// getStaticProps and getServerSideProps
+/**
+ * @type {DrupalState[]}
```
CobyPear, line 29: This is a leftover from some JSDoc typing. Whenever you see these 3 dots under the start of a function definition, its essentially a linting error for the JS/TS doc comment. You can hover over it and it tells you the error and suggests fixes. 

Make sure to update the comments to TSDoc when doing the TS conversion since the JSDoc comments are slightly different (yet another fun JavaScript ecosystem fact 🙃)
Sometimes I use https://tsdoc.org/play/ to make sure my comments display how I expect
Screen Shot 2023-06-22 at 2 58 27 PM

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/package.json.hbs
CobyPear, line 1: Can you write this template in the new format plz?

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  request.headers.forEach((value, key) => (negotiatorHeaders[key] = value))
+
+  // Use negotiator and intl-localematcher to get best locale
+  let languages = new Negotiator({ headers: negotiatorHeaders }).languages()
+  // @ts-ignore locales are readonly
+  const locales: string[] = i18n.locales
+  return matchLocale(languages, locales, i18n.defaultLocale)
+}
+
+export function middleware(request: NextRequest) {
```
CobyPear, line 21: The app-dir-i18n-routing example has some commented out code that we might need:
https://github.com/vercel/next.js/blob/canary/examples/app-dir-i18n-routing/middleware.ts#L24
I think we need to ignore our `/public` files here.
Also, I noticed the `favicon.ico` in the `app/[lang]` dir, is that correct?

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/lib/stores.ts
```diff
@@ -0,0 +1,59 @@
...
+				new DrupalState({
+					apiBase: DRUPAL_URL,
+					debug: debug,
+					clientId: process.env.CLIENT_ID,
+					clientSecret: process.env.CLIENT_SECRET,
+				}),
+		  ];
+
+// exporting so we can reuse in
+// getStaticProps and getServerSideProps
```
CobyPear, line 27: probs can remove this :)

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/lib/constants.tsx
```diff
@@ -0,0 +1,6 @@
+export const DRUPAL_URL = process.env.backendUrl || 'undefined';
+export const IMAGE_URL = process.env.imageUrl || DRUPAL_URL;
+export const BUILD_MODE =
+	process.env.npm_lifecycle_script?.includes('next build');
+export const LOCALES = process.env.locales as unknown as string[];
```
CobyPear, line 5: I don't think we need this anymore, The docs say to fetch the valid locales in the middleware. Not sure that example I sent is doing that either, but we should be able to repurpose the `get-locales` script into the middleware and do away with setting this var at all. In other words, right now we are fetching locales at build time and they are statically available to the app at run time. With the app dir, we should be able to fetch the locale per-request (and I believe the response will be cached with the app dir magic), then the locale can be given to the renderer on the server which spits out the correctly translated content.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type FooterData = {
+	title: string,
+	url: string,
+	parent: string
+}
+
+async function getData(locale: string) {
+	const store = getCurrentLocaleStore(locale, globalDrupalStateStores);
+	const footerMenu = await store.getObject({
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx:21]: Seems like there might be a race condition or something going on here: I am occasionally getting the following error when refreshing the page:

```shell
- error app/[lang]/layout.tsx (21:32) @ getObject
- error TypeError: Cannot read properties of undefined (reading 'getObject')
    at getData (./app/[lang]/layout.tsx:24:36)
    at RootLayout (./app/[lang]/layout.tsx:30:30)
    at async Promise.all (index 0)
  19 | 	const store = getCurrentLocaleStore(locale, globalDrupalStateStores);
  20 | 	console.log(store)
> 21 | 	const footerMenu = await store.getObject({
     | 	                              ^
  22 | 		objectName: 'menu_items--main',
  23 | 	});
  24 |
```

Seems to work smoothly when using optional chaining here which shouldn't be necessary to my understanding.

```suggestion
	const footerMenu = await store?.getObject<FooterData[]>({
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx
```diff
@@ -0,0 +1,73 @@
+import '@pantheon-systems/nextjs-kit/style.css';
+
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx:2]: (very small picky thing) I'm starting to lean towards having 0 separators between imports, but I could be convinced otherwise. Thoughts? Either way I think the CSS imports should be grouped together.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Footer } from '@pantheon-systems/nextjs-kit/dist/components/footer';
+import { Header } from '@pantheon-systems/nextjs-kit/dist/components/header';
+import styles from './layout.module.css';
+import {
+	getCurrentLocaleStore,
+	globalDrupalStateStores,
+} from '../../lib/stores';
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx:9]: The default tsconfig sets up a path alias of `@` for the root dir, and after adding the `src` dir you can import like this:
```suggestion
} from '@/src/lib/stores';
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+async function getData(locale: string) {
+	const store = getCurrentLocaleStore(locale, globalDrupalStateStores);
+	const footerMenu = await store.getObject({
+		objectName: 'menu_items--main',
+		anon: true,
+	});
+
+  return footerMenu as FooterData[];
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/app/[lang]/layout.tsx:26]: No need to typecast, `getObject` supports a generic which should spit the data out as that type (see suggestion above).

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/package.json.hbs
```diff
@@ -21,6 +21,8 @@
 		{{> tailwindcssDeps}}
 		{{/if}}
         "dotenv": "^16.0.2",
+		"@formatjs/intl-localematcher": "^0.4.0",
```
CobyPear, line 24: Something is off with the indentation here. I'm seeing both tabs and spaces in the output
```
		"@pantheon-systems/drupal-kit": "4.6.1",
		"@pantheon-systems/nextjs-kit": "1.7.4-canary.0",
        "dotenv": "^16.0.2",
		"@formatjs/intl-localematcher": "^0.4.0",
		"negotiator": "^0.6.3",
		"next": "^13.4.4",
		"next-seo": "^5.15.0",
		"react": "18.2.0",
		"react-dom": "18.2.0",
		"sharp": "^0.31.3"
	},
	"devDependencies": {
		"@types/negotiator": "0.6.1",
        "@types/node": "20.2.5",
        "@types/react": "18.2.9",
        "@types/react-dom": "18.2.4",
		"@testing-library/react": "13.4.0",
		"@vitejs/plugin-react": "^3.0.1",
```
image

CobyPear (changes requested):
Noticing now that we aren't using the `src` directory. I think we should to align with the other projects now that it is supported. Made a bunch of inline comments too. I'm being a bit picky on the codebase since we have a good opportunity here to really cement good practices and clean code.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/middleware.ts
```diff
@@ -0,0 +1,58 @@
+import { match as matchLocale } from '@formatjs/intl-localematcher';
+import Negotiator from 'negotiator';
+import type { NextRequest } from 'next/server';
+import { NextResponse } from 'next/server';
+import getLocales from '../scripts/get-locales';
```
CobyPear, line 5: Ideally we would get the locales dynamically and wouldn't need to set the at all in the `next.config.js`. Additionally it would be nice if we had no js left and only typescript.

1. Refactor `get-locales.js` into a typescript function in `src/lib` with the same functionality.
2. Remove the locale setting in `next.config.js`
3. in `stores.ts`, use the new function to get the value of locales to pass in to `makeLocaleStores`
4. also update this file with the new function

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/lib/stores.ts
```diff
@@ -0,0 +1,59 @@
...
+	debug = false,
+}: {
+	locales: string[];
+	debug: boolean;
+}) =>
+	locales.length > 1
+		? locales.map(
+				(locale) =>
+					new DrupalState({
+						apiBase: process.env.backendUrl,
```
CobyPear, line 15: Probs need to add this one to the type as well, seeing this error on build:

```shell
Type error: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

  13 | 				(locale) =>
  14 | 					new DrupalState({
> 15 | 						apiBase: process.env.backendUrl,
     | 						^
  16 | 						defaultLocale: locale,
  17 | 						debug: debug,
  18 | 						clientId: process.env.CLIENT_ID,
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/env.d.ts
```diff
@@ -0,0 +1,10 @@
+declare global {
+    namespace NodeJS {
+      interface ProcessEnv {
+        locales: string[];
+        DEBUG_MODE: boolean;
+      }
+    }
+  }
+
+export {}
```
CobyPear, line 10:
```suggestion
export declare global {
	namespace NodeJS {
		interface ProcessEnv {
			locales: string[];
			DEBUG_MODE: boolean;
		}
	}
}

```

remove empty export, add newline, use tabs > spaces

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/package.json.ts
```diff
@@ -8,32 +11,29 @@
...
 	},
 	"dependencies": {
-		"@pantheon-systems/drupal-kit": "{{drupalKitVersion}}",
-		"@pantheon-systems/nextjs-kit": "{{nextjsKitVersion}}",
-		{{#if tailwindcss}}
-		{{> tailwindcssDeps}}
-		{{/if}}
-        "dotenv": "^16.0.2",
+		"@pantheon-systems/drupal-kit": "${String(data.drupalKitVersion)}",
+		"@pantheon-systems/nextjs-kit":  "${String(data.nextjsKitVersion)}",
```
CobyPear, line 22:
```suggestion
		"@pantheon-systems/drupal-kit": "${data.drupalKitVersion}",
		"@pantheon-systems/nextjs-kit":  "${data.nextjsKitVersion}",
```
These should be typed via the `InputIndex` type. You may need to make the config a template file too in order to get around VSCode picking up that tsconfig. Maybe there's some vscode setting we can use to ignore those nested configs instead but I haven't looked into it.

CobyPear (changes requested):
Left some inline comments, I also noticed this go-around that `build + start` does not seem to be working at all :( Any ideas?

```shell
Listening on port 3000 url: http://localhost:3000
Error: connect ECONNREFUSED ::1:55781
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '::1',
  port: 55781
}
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export async function middleware(request: NextRequest) {
+	const pathname = request.nextUrl.pathname;
+	const locales = await getLocales();
+
+	if (
+		[
+			'/favicon.ico',
+			// Your other files in `public`
```
CobyPear, line 32: We also have `collapse.svg` and `pantheon.png`

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+import Negotiator from 'negotiator';
+import type { NextRequest } from 'next/server';
+import { NextResponse } from 'next/server';
+import getLocales from '../scripts/get-locales';
+
+export const i18n = {
+	defaultLocale: 'en',
+} as const;
+
+function getLocale(
```
CobyPear, line 11: Let's default to arrow functions whenever possible.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+	const negotiatorHeaders: Record<string, string> = {};
+	request.headers.forEach((value, key) => (negotiatorHeaders[key] = value));
+
+	// Use negotiator and intl-localematcher to get best locale
+	let languages = new Negotiator({ headers: negotiatorHeaders }).languages();
+	// @ts-ignore locales are readonly
+	return matchLocale(languages, locales, i18n.defaultLocale);
+}
+
+export async function middleware(request: NextRequest) {
```
CobyPear, line 25: Here as well, use an arrow function.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/app/[lang]/layout.tsx
```diff
@@ -0,0 +1,66 @@
...
+import './globals.css';
+import styles from './layout.module.css';
+
+type FooterData = {
+	title: string;
+	url: string;
+	parent: string;
+};
+
+async function getData(locale: string) {
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/app/[lang]/layout.tsx:17]: Use arrow functions when possible

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/src/env.d.ts
mitchellmarkoff: Since this file does not have any import/export statements the empty export is used to convert this to a module and is needed for the types to get recognized in this format
https://stackoverflow.com/questions/45194598/using-process-env-in-typescript

CobyPear: does `export declare global` not work?

mitchellmarkoff: Just adjusted to that and pushed the change 👍

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-app/package.json.ts
```diff
@@ -0,0 +1,52 @@
...
+		"prettier:fix": "prettier '**/*.{js,jsx,,md}' --write --ignore-path .prettierignore",
+		"test": "vitest run",
+		"update-snapshots": "vitest run --update --silent",
+		"coverage": "vitest run --coverage"
+	},
+	"dependencies": {
+		"@pantheon-systems/drupal-kit": "${data.drupalKitVersion}",
+		"@pantheon-systems/nextjs-kit":  "${data.nextjsKitVersion}",
+		"dotenv": "^16.0.2",
+		"@formatjs/intl-localematcher": "^0.4.0",
```
CobyPear, line 25: Small thing, move this to the top of `dependencies` – ideally these should be in alphabetical order.

packages/create-pantheon-decoupled-kit/src/actions/runLint.ts
```diff
@@ -27,6 +27,13 @@ export const runLint: Action = async ({ data }) => {
 			assert: { type: 'json' },
 		});
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+		if (pkg?.scripts['build-scripts']) {
+			execSync(`${command}  build-scripts`, {
+				cwd: data.outDir,
+				stdio: 'inherit',
+			});
+		}
```
CobyPear, line 35: Can you move the `npm run build-scripts` to the `next lint` in the package.json instead? If possible, we should keep the actions generic and move the starter specific logic into the package.json scripts.

CobyPear (changes requested):
Just two small changes and this is good to go.

packages/create-pantheon-decoupled-kit/src/actions/runLint.ts
mitchellmarkoff: Ah yes definitely, that is a more clever way to do it.

CobyPear (approved):

PR #790, opened 2023-07-18
refactor: Use classes for better readability and extensibility

CobyPear:
## What changes were made?
- Remove util functions, instead use class based architecture for easier extensibility
- Use snapshot tests. The unit tests were not really achieving much to be honest. This allows us to test against the console output instead of the unit tests which I did not think were super helpful. Now we can check against the expected terminal output.
- Try to increase readability. There is still probably much work to be done here and I'm not sold on naming but I didn't want to spend a ton of time here.
decoupled-kit-health-check

New tests, also ran it locally
## Additional information
Things could be a bit DRYer still but I think this is a decent start. Each check takes care of the logging and setup of the URL before checking for a 200 and throwing the required error or logging the warning/success depending on the check.
There's probably a much more clever way to abstract things out but I haven't been able figure it out. 

No changeset for the CLI this time since nothing user facing has changed. Will bump that when we release the gatsby+wp check.


mitchellmarkoff (approved):
Tested locally and seems to still be working as normal. LGTM 🕺

PR #794, opened 2023-07-24
DB-6542: [gatsby-wp, gatsby-wp-acf-addon] Use TypeScript for the `gatsby-wp*` templates

CobyPear:
## What changes were made?
This PR pulls in the `feat/typescript-generators` branch with the `gatsby-wp` generators converted to TypeScript
Other notable changes to the CLI:
- `.*.ts` tagged template literal files. These files are written in TS, compiled to JS, then called by the `addWithDiff` action to render the result to a file based on the filename. So `someFile.json.ts` will be `someFile.json` in the final output. This allows us to use tagged template comments like `/* json */` in the codebase to get syntax highlighting and prettier formatting.
- Refactored some types
- Added `taggedTemplateHelpers` which are essentially ports of the handlebars helpers, both built-in and custom.
- Rewrote the `gatsby-wp` and `gatsby-wp-acf-addon` templates to use TypeScript by default. This includes GraphQL generated types using gatsby's graphql codegen. This file may need to be updated in case of any backend changes. It is generated when running `gatsby develop` and should not be edited manually.
- Docs are updated where appropriate
- `copyTemplates.ts` cli script has been updated to account for the new templates

Tested generator locally
## Additional information

Might want to squash merge this one?


mitchellmarkoff (approved):
LGTM! I am really liking the New Way™️ of writing templates so far. There is a big diff here but nothing jumped out to me after a few passovers. Gatsby starter and acf plugin are still working locally as expected so it looks good to go 👍 🕺

PR #795, opened 2023-07-24
DB-6550: [next-wp] Add Search Input to Layout

mitchellmarkoff:
## What changes were made?
- New `next-wp-search-addon-api` generated added
- `<SearchInput />` component added to the next-wp `<Layout />` component.
- Wrote shared templates for the `<SearchInput />` and `<Layout />` components the New Way ™️ 

- `next-wp-search-addon-api`, `next-drupal-search-addon-api`, `next-drupal`, `next-wp`

locally

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/components/layout.jsx
CobyPear, line 1: Given this file is about 2 lines diff from [`./src/templates/next-wp/components/layout.jsx`](https://github.com/pantheon-systems/decoupled-kit-js/blob/canary/packages/create-pantheon-decoupled-kit/src/templates/next-wp/components/layout.jsx), can we make it a template? Any time 2 files are going to share a good amount of code, it should probably be a template. Especially for these add-ons where we don't want to have to make changes in 2 places if we update the base `Layout`

packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/components/search-input.jsx
CobyPear, line 1: Same here – if this component isn't any different from the one in `next-drupal-search-api-addon`, we should probably consider moving it to `nextjs-kit` or at least make it shareable between the add-ons.

packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/components/searchInput.module.css
CobyPear, line 1: Same here :)

CobyPear (changes requested):
Embrace the templates!

CobyPear:
It might be worth waiting for #794 to be merged so you can write the templates in the New Way™️

packages/create-pantheon-decoupled-kit/src/templates/partials/nextjs-shared/searchInputCSSModulesT.ts
```diff
@@ -0,0 +1,56 @@
+export const searchInputCSSModules = () => `.container {
```
CobyPear, line 1:
```suggestion
export const searchInputCSSModules = () => /* css */ `.container {
```
The lang comments (idk what to call these?) can be added to the partials as well.

packages/create-pantheon-decoupled-kit/src/templates/tailwindless-next/components/layout.module.css.ts
```diff
@@ -0,0 +1,10 @@
+import { layoutCSSModules } from '@cli/templates/partials/nextjs-shared/layoutCSSModulesT';
+import { TemplateFn } from '@cli/types';
+
+const css: TemplateFn = ({ data, utils }) =>
+	/* css */ `${utils.if(data.search, layoutCSSModules(true))} ${utils.if(
+		!data.search,
+		layoutCSSModules(false),
+	)}`;
```
CobyPear, line 8:
```suggestion
	/* css */ `${layoutCSSModules(data.search)}`;
```

packages/create-pantheon-decoupled-kit/src/templates/partials/nextjs-shared/layoutT.ts
CobyPear, line 1: All templates should have tagged template comment like `/* jsx */`

packages/create-pantheon-decoupled-kit/src/templates/partials/nextjs-shared/layoutT.ts
```diff
@@ -0,0 +1,95 @@
...
+					<a
+						href="https://www.drupal.org/"
+					>
+						Drupal
+					</a>`;
+
+export const layoutTemplate = (
+	search: boolean,
+	cmsType: string,
+) => /* jsx */ `import { Footer, Header, PreviewRibbon } from '@pantheon-systems/nextjs-kit';
```
CobyPear, line 71: Use named args if there is more than one, this way it's easier to tell what these args mean when we're calling the templates
```suggestion
export const layoutTemplate = ({ search, cmsType } :{
	search: boolean,
	cmsType: string,
}) => /* jsx */ `import { Footer, Header, PreviewRibbon } from '@pantheon-systems/nextjs-kit';
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal/components/layout.jsx.ts
```diff
@@ -0,0 +1,6 @@
+import { layoutTemplate } from '@cli/templates/partials/nextjs-shared/layoutT';
+import { TemplateFn } from '@cli/types';
+
+const jsx: TemplateFn = () => /* jsx */ `${layoutTemplate(false, 'drupal')}`;
```
CobyPear, line 4:
```suggestion
const jsx: TemplateFn = ({ data }) => /* jsx */ `${layoutTemplate({search: data.search, cmsType: data.drupal || data.wp })}`;
```
Use the `data` instead of explicit args. Also use named args on the templates if there is > 1

packages/create-pantheon-decoupled-kit/src/generators/next-wp-search-api-addon.generator.ts
```diff
@@ -0,0 +1,40 @@
...
+	convertCSSModules,
+	runInstall,
+	runLint,
+} from '../actions';
+import type { DecoupledKitGenerator, DefaultAnswers } from '../types';
+import { outDirPrompt, tailwindcssPrompt } from '../utils/sharedPrompts';
+
+interface NextWpSearchApiAddonAnswers {
+	search: true;
+}
```
CobyPear, line 13: I moved these interfaces into `types.ts` and made a `BaseGeneratorData` type that these should extend. This way all of the data passed into the templates is type-safe. LMK if you want a more detailed walkthrough on that.

CobyPear (changes requested):
Looking much better from my chaotic perspective! I can tell you are grasping the power of these templates/partials :)

A few general notes in addition to the inline comments:

The `T` at the end of the file name isn't necessary. I used it for some of the other templates because I didn't want to replace the handlebars template of the same name but wanted a way to distinguish between the two; in hindsight that was probably more confusing than helpful. So I would remove that from your new templates, and I plan to remove the `T` when we refactor the other starters and remove the handlebars templates all together.

One thing that I could have probably documented better: The partials can take arbitrary arguments. If there is >1 I would suggest making them named arguments, possibly even with defaults either way.
When calling these partials in the templates, use the `data` prop to get these values so that the CLI can pass in args from the user into these spots. This should hopefully make things a bit easier to read as well.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal/components/layout.jsx.ts
CobyPear: Err actually the `data.drupal || data.wp` won't really work since those are booleans and we aren't passing those as strings anywhere. Maybe there's some trick we can do with the `CMSType` and our `Input` type.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/components/layout.jsx.ts
```diff
@@ -0,0 +1,10 @@
+import { layoutTemplate } from '@cli/templates/partials/nextjs-shared/layout';
+import { TemplateFn } from '@cli/types';
+
+const jsx: TemplateFn = ({ data }) =>
+	/* jsx */ `${layoutTemplate({
+		search: data.search,
+		cmsType: data.cmsType,
```
CobyPear, line 7: We're not currently passing `cmsType` in as `data`, but we could...

packages/create-pantheon-decoupled-kit/src/generators/next-wp-search-api-addon.generator.ts
```diff
@@ -0,0 +1,38 @@
...
+	runLint,
+} from '../actions';
+import type { BaseGeneratorData, DecoupledKitGenerator } from '../types';
+import { outDirPrompt, tailwindcssPrompt } from '../utils/sharedPrompts';
+
+interface NextWpSearchApiAddonAnswers extends BaseGeneratorData {
+	search: boolean;
+}
+
+export const nextWpSearchApiAddon: DecoupledKitGenerator<NextWpSearchApiAddonAnswers> =
```
CobyPear, line 15: This should be in `types.ts` and the name should be `Data` not `Answers`. The base answers still need to be passed in to the generic type.

Also, it should be moved to `types.ts` so that it can be added to `InputIndex`. Right now `Input` aka `data` doesn't know about the `search` or `cmsType` prop.

CobyPear (changes requested):

packages/create-pantheon-decoupled-kit/src/templates/next-drupal/components/layout.jsx.ts
```diff
@@ -0,0 +1,10 @@
+import { TemplateFn, isDrupalCms } from '@cli/types';
+import { layoutTemplate } from '@partials/nextjs-shared/layout';
+
+const jsx: TemplateFn = ({ data }) =>
+	/* jsx */ `${layoutTemplate({
+		search: data.search,
+		cmsType: isDrupalCms(data.cmsType) ? 'drupal' : 'false',
```
CobyPear, line 7: I think it's fine to keep this as `data.cmsType`, but you can use the type guard in the `layoutTemplate`
```suggestion
		cmsType: data.cmsType,
```

packages/create-pantheon-decoupled-kit/src/templates/partials/nextjs-shared/layout.ts
```diff
@@ -0,0 +1,102 @@
...
+				<SearchInput />
+			</div>`,
+			)}
+			${utils.if(!search, `<Header navItems={navItems} />`)}
+			<main className={styles.layoutMain}>{children}</main>
+			<Footer footerMenuItems={footerMenuItems}>
+				<span className={styles.footerCopy}>
+					© {new Date().getFullYear()} Built with{' '}
+                    ${utils.if(cmsType === 'wp', wpSig)}
+                    ${utils.if(cmsType === 'drupal', drupalSig)}
```
CobyPear, line 96:
```suggestion
                    ${utils.if(isWpCMS(cmsType), wpSig)}
                    ${utils.if(isDrupalCms(cmsType), drupalSig)}
```

CobyPear (changes requested):

packages/create-pantheon-decoupled-kit/src/templates/partials/nextjs-shared/layout.ts
mitchellmarkoff: Ooo you're right that is def cleaner.

CobyPear (approved):
Nice job! 👯

PR #807, opened 2023-07-27
DB-6549: [next-wp-search-api-addon] Search Results Page

mitchellmarkoff:
## What changes were made?
- Small PR adding a search results page to the next-wp-search-api-addon. Right now the page just renders a header 'Search Results'

cli < templates < next-wp-search-api-addon

locally

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/[[...searchAlias]].jsx
```diff
@@ -0,0 +1,26 @@
+import { NextSeo } from 'next-seo';
```
Comment (mitchellmarkoff) [inline packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/[[...searchAlias]].jsx:1]: I think for now it makes sense to have this search page not be a shared template with the drupal search addon. Since we will be fetching our search result data from Pantheon Solr, we will be getting much differently shaped data to de-structure and display, along with fetching it differently. Once all of that is implemented it may make sense to combine into one template if there are enough similarities but I'm not sure there will be yet.

CobyPear (approved):

PR #824, opened 2023-08-10
DB-5859: Refactor branding and navigation

CobyPear:
## What changes were made?
The branding actually did not change that much. Since Pantheon does not have a distinct logo or branding for Open Source projects at the moment, it was determined that there was not a ton to change here. Not too late to change things of course but I just wanted to point out that this didn't feel too different branding wise.

- The landing page is gone and now redirects straight to `/docs` which has the overview page.
- Created an index page for each category heading which allows for another mode of navigation opposed to the sidebar
- SEARCH! found a very easy way to add search. There is another plugin that is a bit more configurable as to what text is weighted but this one seemed to be somewhat more up to date. (Also due to a user error I couldn't get the first one working and didn't go back to it once I got the current implantation working. If we need to fine tune stuff we can change it.)
    - note: the search component only works in production mode, not dev mode. See the note in the README.
- Added the `<DocCardList />` to the Overview page as well. This required [swizzling](https://docusaurus.io/docs/swizzling) to eject the required components from the theme and then refactoring them to work outside of the `context` that was needed for those components. Seems to work fine and since I renamed the swizzled components, they do not clash with the other `DocCard` and `DocCardList` components. So they're not technically swizzled any more because the original implementation is still working and being used for the other index pages, but it is fun to say. One downside is that if these components change upstream, we may need to incorporate those changes to the custom components.

docs

## Additional information


backlineint (approved):
Card navigation and search are both great enhancements, and landing on the overview page seems much more efficient.

PR #827, opened 2023-08-17
DB-6233: Add comments to .env.example env vars

CobyPear:
## What changes were made?
Added some explanation comments to the .env.examples
all 3 starters

n/a
## Additional information

### DIFF
--- a/packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
+++ b/packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
@@ -1,13 +1,15 @@
 # Copy as .env.development.local to override envars for local development
+# Optional environment variables are commented out
 
 # The graphql endpoint of your WordPress instance
 WPGRAPHQL_URL=https://my-backend-site.pantheonsite.io/wp/graphql
 
 # Username and Application Password for a valid user in the WordPress instance
-WP_APPLICATION_USERNAME=user
-WP_APPLICATION_PASSWORD=example_pass
+# Required to fetch private content from WordPress
+# WP_APPLICATION_USERNAME=user
+# WP_APPLICATION_PASSWORD=example_pass
 
-# The value of Gatsby pathPrefix
-# See https://www.gatsbyjs.com/docs/how-to/previews-deploys-hosting/path-prefix/ for more information
+# Sets the gatsby.config.pathPrefix setting for you
 # This value is only required if using the pathPrefix feature
-PANTHEON_UPLOAD_PATH=/example
+# See https://www.gatsbyjs.com/docs/how-to/previews-deploys-hosting/path-prefix/ for more information
+# PANTHEON_UPLOAD_PATH=/example
--- a/packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
+++ b/packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
@@ -1,7 +1,9 @@
+# Copy as .env.development.local to override envars for local development
+
 # The graphql endpoint of your WordPress instance
 WPGRAPHQL_URL=https://my-backend-site.pantheonsite.io/wp/graphql
 
-# Credentials for a valid user in the WordPress instance
+# Username and Application Password for a valid user in the WordPress instance
 WP_APPLICATION_USERNAME=user
 WP_APPLICATION_PASSWORD=example_pass
 
--- a/packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
+++ b/packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
@@ -1,5 +1,11 @@
-# Optional environment variables are commented out
+# The graphql endpoint of your WordPress instance
 WPGRAPHQL_URL=https://my-backend-site.pantheonsite.io/wp/graphql
-#PANTHEON_UPLOAD_PATH=/example
-#WP_APPLICATION_USERNAME=
-#WP_APPLICATION_PASSWORD=
+
+# Credentials for a valid user in the WordPress instance
+WP_APPLICATION_USERNAME=user
+WP_APPLICATION_PASSWORD=example_pass
+
+# The value of Gatsby pathPrefix
+# See https://www.gatsbyjs.com/docs/how-to/previews-deploys-hosting/path-prefix/ for more information
+# This value is only required if using the pathPrefix feature
+PANTHEON_UPLOAD_PATH=/example
--- a/packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
+++ b/packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
@@ -10,7 +10,7 @@ BACKEND_URL=http://my-backend-site.pantheonsite.io
 # IMAGE_DOMAIN=some-other-domain.example
 
 # The secret set on your Drupal Preview Site. Required for preview functionality
-# To set a new secret, go to https://{your Drupal domain}/admin/structure/dp-preview-site
+# See https://decoupledkit.pantheon.io/docs/backend-starters/decoupled-drupal/configuring-preview-site for more information on configuring a preview site
 # PREVIEW_SECRET=super-secret-preview-secret
 
 # The credentials of your Drupal OAuth client
--- a/packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
+++ b/packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
@@ -1,14 +1,23 @@
-# Copy as .env.development.local to override envars for local development
+# The URL of your Drupal site
 BACKEND_URL=http://my-backend-site.pantheonsite.io
-FRONTEND_URL=http://my-frontend-site.pantheonsite.io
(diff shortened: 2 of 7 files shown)


packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
```diff
@@ -1,14 +1,23 @@
-# Copy as .env.development.local to override envars for local development
+# The URL of your Drupal site
 BACKEND_URL=http://my-backend-site.pantheonsite.io
-FRONTEND_URL=http://my-frontend-site.pantheonsite.io
-IMAGE_DOMAIN=my-frontend-site.pantheonsite.io
-PREVIEW_SECRET=mysecret
+
+# The domain where your images are hosted. By default, we will set the domain of the BACKEND_URL
+# to the value of IMAGE_DOMAIN. Set this if your images are hosted somewhere other than your WordPress instance
```
backlineint, line 5: Some instances of WordPress here that need to be changed to Drupal.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
```diff
@@ -1,14 +1,23 @@
-# Copy as .env.development.local to override envars for local development
```
backlineint, line 1: Could you elaborate on the reasoning behind removing this? I feel like guidance like this is helpful and probably should be in the other example env files.

packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
```diff
@@ -1,5 +1,11 @@
-# Optional environment variables are commented out
+# The graphql endpoint of your WordPress instance
 WPGRAPHQL_URL=https://my-backend-site.pantheonsite.io/wp/graphql
-#PANTHEON_UPLOAD_PATH=/example
-#WP_APPLICATION_USERNAME=
-#WP_APPLICATION_PASSWORD=
+
+# Credentials for a valid user in the WordPress instance
+WP_APPLICATION_USERNAME=user
```
backlineint, line 5: Kinda thinking out loud here, but my concern is the username and password for the account. The application password is something different.

Would something like this make it better, or is it still confusing?

`# Username and Application Password for a valid user in the WordPress instance`

backlineint (changes requested):
A few minor things, but this is an improvement from my perspective.

@stevector is there anything else you were hoping to see here. You submitted the original bug that inspired this.

packages/create-pantheon-decoupled-kit/src/templates/gatsby-wp/.env.example
CobyPear: Makes more sense to me, I forget this distinction myself quite often.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
CobyPear: No good reason, I will add it back.

packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
CobyPear: Oops!

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
```diff
@@ -1,7 +1,21 @@
...
-#PANTHEON_UPLOAD_PATH=/example
-#PREVIEW_SECRET=
-#WP_APPLICATION_USERNAME=
-#WP_APPLICATION_PASSWORD=
+# Copy as .env.development.local to override envars for local development
+
+# The graphql endpoint of your WordPress instance
+WPGRAPHQL_URL=https://my-backend-site.pantheonsite.io/wp/graphql
+
+# The secret set on your WordPress Preview Site. Required for preview functionality
```
stevector, line 6: How/Where do I (the user/developer) find out what this secret is? Can we add a link to where I would find it?

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
```diff
@@ -1,7 +1,21 @@
...
+PREVIEW_SECRET=super-secret-preview-secret
+
+# The domain where your images are hosted. By default, we will set the domain of the WPGRAPHQL_URL
+# to the value of IMAGE_DOMAIN. Set this if your images are hosted somewhere other than your WordPress instance
+# See https://nextjs.org/docs/pages/api-reference/components/image#domains for more information
+IMAGE_DOMAIN=some-other-domain.example
```
stevector, line 12: I think the convention used previously was good

```
# Optional environment variables are commented out
```

Having this variable be not commented out makes me think I need it. And if someone does blindly copy the whole file, and start changing values, they'll end up with broken images as long as they keep `IMAGE_DOMAIN=some-other-domain.example` right?

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
stevector: Also, please explain "preview" more. Maybe something like


> This variable is necessary when content editors use preview functionality in the WordPress interface to see/preview a content change before that content change becomes public. This front-end application uses this secret to retrieve the not-yet-public content from WordPress. To find this value....

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
```diff
@@ -1,7 +1,21 @@
...
+
+# Username and Application Password for a valid user in the WordPress instance
+WP_APPLICATION_USERNAME=user
```
stevector, line 15: Can we explain why the front-end application would need this info?

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
```diff
@@ -1,7 +1,21 @@
...
+WP_APPLICATION_PASSWORD=example_pass
+
+# The value of Next.js basePath.
+# See https://nextjs.org/docs/pages/api-reference/next-config-js/basePath for more information
+# This variable is only required if using the basePath feature
+PANTHEON_UPLOAD_PATH=/example
```
stevector, line 21: How about 

> If you are using Next.js' `basePath` setting, then also set `PANTHEON_UPLOAD_PATH` to the same value. Setting `PANTHEON_UPLOAD_PATH` will [ok, I don't know what it'll do, but we should say what it does]. If you are not setting Next.js' `basePath`  then you probably don't need to set `PANTHEON_UPLOAD_PATH`.

stevector:
@backlineint thanks! I dropped in some thoughts as line-by-line comments.

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
CobyPear: > Also, please explain "preview" more. Maybe something like ...

Would a link to our docs suffice?
https://decoupledkit.pantheon.io/docs/frontend-starters/nextjs/nextjs-drupal/implementing-preview

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
CobyPear: This var actually sets the `basePath`/`pathPrefix` for the user. how about:

```
# Sets the next.config.basePath setting for you
# This variable is only required if using the basePath feature
# See https://nextjs.org/docs/pages/api-reference/next-config-js/basePath for more information
```

packages/create-pantheon-decoupled-kit/src/templates/next-drupal/.env.example
```diff
@@ -1,14 +1,27 @@
 # Copy as .env.development.local to override envars for local development
+# Optional environment variables are commented out
+
+# The URL of your Drupal site
 BACKEND_URL=http://my-backend-site.pantheonsite.io
-FRONTEND_URL=http://my-frontend-site.pantheonsite.io
```
backlineint, line 3: FRONTEND_URL is probably not commonly used, but it is still something that we honor in next.config.js, so I wonder if we shouldn't remove this. 

I actually has to look it up in our docs to remember what it does...

---

If your site is translated and you would like the hreflang metadata set correctly, you may set FRONTEND_URL to the URL of your frontend site.

If the FRONTEND_URL is not set, it will default to the value of PANTHEON_ENVIRONMENT_URL

---

My assumption is that this would still be necessary for translated production sites that have a live domain - I'm assuming that the PANTHEON_ENVIRONMENT_URL is the platform URL, but that might be worth revisiting.

CobyPear: It is actually set automatically in the next.config! We could still add it here to call it out, but I don't mind "hiding" this one.

packages/create-pantheon-decoupled-kit/src/templates/next-wp/.env.example
backlineint: Personally I prefer a link to documentation to a paragraph of explanation. Seems to verbose for this file.

If we add that, we'd need to use the Drupal specific link here.

PR #829, opened 2023-08-21
DB-6547: Functioning Search Route

mitchellmarkoff:
## What changes were made?
- A query will be executed when a user submits a search and all results will be shown dynamically in on the search page.
- To aid this work a template was made for the `Posts.js` file used by the nextjs starters. This brought some small housekeeping updates to the `next-wp` and `next-wp-acf-addon` templates.

`next-wp`, `next-wp-acf-addon`, `next-wp-search-api-addon`

locally

## Additional information


packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/[[...searchAlias]].jsx
```diff
@@ -1,26 +1,95 @@
 import { NextSeo } from 'next-seo';
```
Comment (mitchellmarkoff) [inline packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/[[...searchAlias]].jsx:1]: There is probably an argument that could be made to make this a shared template with the search page in the `next-drupal` starter. I chose not to because I feel like the logic in `getServerSideProps` as well as how the data shapes are destructured both differ enough from one another to warrant their own files. They do still share some bits though. The `next-drupal` file is [here](https://github.com/pantheon-systems/decoupled-kit-js/blob/feat/wp-search-addon/packages/create-pantheon-decoupled-kit/src/templates/next-drupal-search-api-addon/pages/search/%5B%5B...searchAlias%5D%5D.jsx) if you want to compare the two and weigh in any thoughts.

Comment (CobyPear) [reply packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/[[...searchAlias]].jsx:1]: > I chose not to because I feel like the logic in getServerSideProps as well as how the data shapes are destructured both differ enough from one another to warrant their own files. 

Sounds reasonable to me.

packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/searchPage.module.css.ts
```diff
@@ -0,0 +1,6 @@
+import { TemplateFn } from '@cli/types';
+import { searchPageCSSModules } from '@partials/nextjs-shared/searchPageCSSModules';
+
+const css: TemplateFn = () => `${searchPageCSSModules()}`;
```
CobyPear, line 4:
```suggestion
const css: TemplateFn = () => searchPageCSSModules();
```

The `${}` is somewhat redundant here.

packages/create-pantheon-decoupled-kit/src/templates/partials/next-wp/Posts.ts
CobyPear, line 1: if these are all meant to be queires, can you name them as such? `postUriAcf` doesn't really mean anything to me. What is this query for?

I think these partials could be broken up a bit more to avoid using the backticks util which removes the syntax highlighting for the graphql queries. See the partial at `partials/gatsby-wp/wpPostsQuery.ts`: we can put the query itself in a variable and then insert this query to the function inside the template.

I might even seperate out the queries to their own file like `wpPostsQueries` then import the partials into `wpPostLib`, so the partials in that file would be a bit cleaner like so:

```ts
import { wpSearchQuery } from '@partials/next-wp/wpPostsQueries';
// ..
const searchQuery = /* js */ `
export async function getSearchedPosts(searchTerm) {
	const query = gql${utils.backticks(wpSearchQuery())}

	const {
		data: {
			posts: { edges },
		},
		headers,
	} = await client.rawRequest(query, { searchTerm });
	const posts = edges.map(({ node }) => node);
	return { posts, headers };
}
`;

```

To me this is a lot easier to read, and we can maintain the GraphQL syntax highlighting in the Query partial.

Also can you rename this file to camel case and be more descriptive such as `wpPostsQueries`.

CobyPear (changes requested):
Functionally it is working 🥳 
I think we can clean up some of these partials a bit tho, see comments.

packages/create-pantheon-decoupled-kit/src/templates/partials/next-wp/Posts.ts
```diff
@@ -0,0 +1,167 @@
...
+	} = await client.rawRequest(query, { uriString });
+
+	return { post, headers };
+}
+`;
+const searchQuery = /* js */ `
+export async function getSearchedPosts(searchTerm) {
+	const query = gql ${utils.backticks`
+		query LatestPostsQuery($searchTerm: String) {
+			posts(where: { search: $searchTerm }) {
```
backlineint, line 77: @mitchellmarkoff @CobyPear this query looks pretty much identical to what the query would be if you weren't using Solr. It is also referencing posts rather than an index. Did you confirm that this is actually taking advantage of the Solr index you configured? A couple of ways you could do that:

* Assuming that the index you created indexes both posts and pages (the acceptance criteria in the ticket specifies that this search should return posts and pages) try searching for something that should match pages but not posts. If you get no results, this might not be using the index.
* Alternatively, point this to a WP instance that doesn't have Solr configured. if you still get results, that indicates that this isn't using Solr at all.

My best guess is that this is using default WordPress search, but maybe the Solr plugin does some magic I'm not aware of (the docs do say things like 'Completely replaces default WordPress search'.) Didn't mean to lead you in the wrong direction when asking if we could get at Solr data via GraphQL.

CobyPear: Oh good call. I'm not seeing any Pages come back from the search.

packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/[[...searchAlias]].jsx
```diff
@@ -1,26 +1,95 @@
...
+			{error ? (
+				<div className={styles.altResult}>
+					<span>An error occurred while fetching search results</span>
+				</div>
+			) : (
+				<section className={styles.section}>
+					<div className={styles.container}>
+						{searchResults?.length > 0 ? (
+							<ul>
+								{searchResults?.map(({ title, excerpt, uri, postId }) => (
```
Comment (CobyPear) [inline packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/pages/search/[[...searchAlias]].jsx:31]: Forgot to mention this on the review: There could be a bit more space between the results, looks a bit crowded right now.
Screen Shot 2023-08-22 at 9 13 28 AM

packages/create-pantheon-decoupled-kit/src/templates/partials/next-wp/Posts.ts
mitchellmarkoff: @backlineint  Here is what I had done to test that Solr is being used here, I used two CMS sites. One without Solr enabled and one with Solr.

Each of these CMS sites have 3 different posts titled `A Example`, `B Example`, and `C Example`. These posts were created in the order I just listed them.

Expectations when querying the term `example`:
- For the CMS without Solr, I would expect to get results from the default WP search that are ordered by the date/time they were created, with the newest post being shown first. Expected order - `C Example`, `B Example`, `A Example`.
- For the CMS with Solr, I would expect the results to come back based on score, which in this case should place a higher emphasis on alphabetical order since the search term is present in the same place of each of these posts. Expected order - `A Example`, `B Example`, and `C Example`

When I went to confirm this I got results that matched the expectations above and lead me to believe that Solr is working as expected here.

Feel free to checkout the test sites I used:
Without Solr: https://dev-gql-search.pantheonsite.io/wp-admin/admin.php?page=graphiql-ide
With Solr: https://dev-wp-search-demo.pantheonsite.io/wp-admin/admin.php?page=graphiql-ide
Normal WP QA credentials for both.

Query used in each:
```
query NewQuery {
  posts(where: {search: "example"}) {
    nodes {
      title
      date
    }
  }
}
```

Results from site without Solr:
```
{
  "data": {
    "posts": {
      "nodes": [
        {
          "title": "C Example",
          "date": "2023-08-22T15:40:42"
        },
        {
          "title": "B Example",
          "date": "2023-08-22T15:39:53"
        },
        {
          "title": "A Example",
          "date": "2023-08-22T15:39:37"
        }
      ]
    }
  },
```

Results from site with Solr:
```
{
  "data": {
    "posts": {
      "nodes": [
        {
          "title": "A Example",
          "date": "2023-08-22T15:41:23"
        },
        {
          "title": "B Example",
          "date": "2023-08-22T15:41:34"
        },
        {
          "title": "C Example",
          "date": "2023-08-22T15:41:43"
        }
      ]
    }
  },
```

packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/lib/Search.js
```diff
@@ -0,0 +1,52 @@
+import { gql } from '@pantheon-systems/wordpress-kit';
+import { client } from './WordPressClient';
+
+export async function getSearchResults(searchTerm) {
```
mitchellmarkoff, line 4: Some context on what is happening with this query:

When accessing the Solr index via graphql, we are confined to the fields exposed in our graphql schema. Since our does not offer a way to hit both posts and pages in the same place, I have two separate statements in the query below. One for posts and one for pages, I then concatenate the results together for the return value.

While there are some extra steps here, the response object keeps it's correct scored and ordered shape. Solr scores pages below posts. Because of this, any post that contains a match to the search term should (and does) show before the highest ranked page.

This query (not using graphql) shows how posts will come before pages in scoring when looking at the Solr index, what I have below is replicating the same behavior. https://dev-wp-search-demo.pantheonsite.io//wp-json/wp/v2/search?search=example

Since we are bringing pages into the mix now I decided to break this out into it's own `Search.js` file as I don't think `Posts.js` suits it anymore.

packages/create-pantheon-decoupled-kit/src/templates/partials/next-wp/Posts.ts
mitchellmarkoff: Implemented your suggestions above, really enjoying working with these templates the more I get into them, couldn't imagine going back to handlebars after working with these😅

As I mentioned above I broke the search query into it's own `Search.js` file, because of that the search query no longer lives here and this template is not needed as a part of the work for this PR but keeping as it still eliminates an old `hbs` template. This template is handling the addition of the ACF fields to the `getPostByUri ` query if needed.

packages/create-pantheon-decoupled-kit/src/templates/next-wp-search-api-addon/lib/Search.js
CobyPear: Thanks for the explainer, seems good to me!

packages/create-pantheon-decoupled-kit/src/templates/partials/nextjs-shared/searchPageCSSModules.ts
```diff
@@ -0,0 +1,44 @@
...
+	margin-top: var(--4);
+	justify-items: start;
+}
+
+@media (min-width: 1024px) {
+	.section {
+		min-width: var(--md);
+	}
+	.container {
+		max-width: var(--lg);
```
CobyPear, line 41: I don't think we should change anything now, but I'm noticing when I search "hello" and get the single "hello world" post back, the width of the container is `sm`, but with a search like "example", since the excerpts are longer it makes the container width `md` or `lg`. So going between the two searches is a bit jarring. I think we should address this in our future re-theme work.

packages/create-pantheon-decoupled-kit/src/templates/partials/next-wp/Posts.ts
CobyPear: > really enjoying working with these templates the more I get into them, couldn't imagine going back to handlebars after working with t

That is super great to hear! Once we're able to lint them properly I think they will be even better!!

> I broke the search query into it's own Search.js file

Makes sense to me, this looks much cleaner now

CobyPear (approved):
Nice job on this!

PR #861, opened 2023-09-11
DB-4833: specify platform docs

CobyPear:
## What changes were made?
- Added a shiny admonition to call out Pantheon specific documentation. I think there is a reality where we need to be able to mention the platform and our docs cannot be 100% agnostic as much as we'd like them to be. This will help call out these docs/sections so that it is clear that they are intended for use in that context.
- revised a few docs to use the new admonition.
- Added git clone instructions for the decoupled upstreams on the 'creating a new drupal/wp project' docs.

docs. Swizzled the Admonitions to add the new one. Added the relevant custom CSS as well.
ran docs site locally
## Additional information
This guide was very helpful especially for the CSS part: https://docusaurus.community/knowledge/component-library/existing/Admonitions/

### DIFF
--- a/web/docs/Backend Starters/Decoupled Drupal/creating-new-project.mdx
+++ b/web/docs/Backend Starters/Decoupled Drupal/creating-new-project.mdx
@@ -5,6 +5,37 @@ slug: '/backend-starters/decoupled-drupal/creating-a-new-project'
 sidebar_position: 0
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+## Clone the Repository
+
+To start a new project, clone one of the following repositories:
+
+<Tabs>
+  <TabItem value="drupal10" label="Drupal 10">
+
+```bash
+git clone git@github.com:pantheon-upstreams/decoupled-drupal-10-composer-managed.git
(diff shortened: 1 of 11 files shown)


web/src/theme/Admonition/index.js
CobyPear, line 1: This component was ejected from the default theme via [swizzling](https://docusaurus.io/docs/swizzling). Maybe we can use a wrapper instead but I was having trouble with that at first.

web/docs/Backend Starters/Decoupled Drupal/creating-new-project.mdx
```diff
@@ -5,6 +5,37 @@ slug: '/backend-starters/decoupled-drupal/creating-a-new-project'
...
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+## Clone the Repository
+
+To start a new project, clone one of the following repositories:
+
+<Tabs>
```
backlineint, line 15: Tabs are nice here :)

backlineint (approved):
This does the job nicely and gives us a helpful pattern for the future. Thanks @CobyPear!

PR #868, opened 2023-09-18
DB-6721: fix API reference generation

CobyPear:
## What changes were made?
Way more changes than I would've liked or expected but I kind of just wanted to get things into a working state. Nothing too drastic has changed, but there are a lot of them (85 files, a lot of these are from the documentation generation itself - should that be its own PR?).

Some of the things blocking the typedoc plugin was that it could not compile the typescript because of some type errors in the CLI. These have been fixed but feels a bit hacky. I figured we would revisit some of these types before we deem the CLI ready for 1.0 anyway, so I'm not too concerned about any perceived mess right now. Stricter types on Data passed into the actions meant some of the tests needed to be refactored as well. 

A shared data object was added and a few of the action tests were refactored to use it. In the future I think these tests will be better off as snapshot tests like we did for the health-checks. I will open a ticket for this as I think it will make maintaining/reading these tests a lot easier. That also brings me to another small change: the vitest version is now the up to date across all package except for the CLI where I was seeing a large amount of tests failing for an unknown reason. Maybe we can roll the vitest upgrade into the conversion to snapshot tests for the CLI. Other packages are fine with the latest vitest.

Slight refactoring was done to the way the typedoc options are generated. I think this makes the docusaurus.config.js a bit easier to read but open to changing anything here.

Fixed the broken link warnings from docusaurus as well.

CLI types and tests, docs, packages (vitest upgrade)

## Additional information


backlineint (approved):
Thanks for working through all of this @CobyPear

PR #877, opened 2023-09-25
DB-5856: [docs] Audit uses of back-end and front-end for consistency

backlineint:
## What changes were made?

Changed to use the term 'backend' in all cases.
Use the term 'frontend' unless the context is the Front-end Sites product.

There was some Slack discussion if this is the correct language long term. I think this change still makes sense as it is small in scope, provides consistency, and doesn't change any URL slugs. We can change again if necessary, and we can account for having to handle redirects and updating dependent links in the estimate.

Docs md/mdx files.

Reviewing docs site locally.

## Additional information

### DIFF
--- a/web/docs/Backend Starters/Decoupled Drupal/configuring-existing-project.md
+++ b/web/docs/Backend Starters/Decoupled Drupal/configuring-existing-project.md
@@ -6,10 +6,10 @@ sidebar_position: 1
 ---
 
 While we offer a [backend starter project](./creating-a-new-project) to simplify
-the process of configuring a Drupal site for use with our front-end starter
-kits, you may instead prefer to use an existing Drupal project. Follow the steps
-below to configure an existing Drupal project to work with one of our front-end
-starter kits.
+the process of configuring a Drupal site for use with our frontend starter kits,
+you may instead prefer to use an existing Drupal project. Follow the steps below
+to configure an existing Drupal project to work with one of our frontend starter
+kits.
 
 ## Before You Begin
 
@@ -17,7 +17,7 @@ These instructions assume that you have already installed Drupal using your
 preferred method.
 
 The amount of necessary configuration will vary depending on the features you
-intend to use within your front-end starter kit. As a result, the instructions
+intend to use within your frontend starter kit. As a result, the instructions
 below are broken down into three related sections.
 
 ## 1. Configuring Basic Builds
@@ -68,10 +68,10 @@ Our starter kits assume that there is at least one published article and page in
 your Drupal backend. If your site does not have any article or page content, you
 should create some before proceeding.
 
-### Set the Necessary Front-End Environment Variables
+### Set the Necessary Frontend Environment Variables
 
 At this point, your Drupal site should be configured to allow data to be sourced
-anonymously by the front-end starter kit. Within your front-end project you will
+anonymously by the frontend starter kit. Within your frontend project you will
 also need to set the necessary environment variables to source data from your
 Drupal backend. For anonymous data sourcing you will need to set at least the
 `BACKEND_URL` and `IMAGE_DOMAIN` variables.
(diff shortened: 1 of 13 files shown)


web/docs/Backend Starters/Decoupled Drupal/configuring-existing-project.md
```diff
@@ -5,19 +5,19 @@ slug: '/backend-starters/decoupled-drupal/configuring-an-existing-project'
 sidebar_position: 1
 ---
 
-While we offer a [back-end starter project](./creating-a-new-project) to
-simplify the process of configuring a Drupal site for use with our front-end
-starter kits, you may instead prefer to use an existing Drupal project. Follow
-the steps below to configure an existing Drupal project to work with one of our
-front-end starter kits.
+While we offer a [backend starter project](./creating-a-new-project) to simplify
+the process of configuring a Drupal site for use with our frontend starter kits,
```
CobyPear, line 9:
```suggestion
the process of configuring a Drupal site for use with our front-end starter kits,
```
Used as adjective

web/docs/Backend Starters/Decoupled Drupal/local-development.md
```diff
@@ -12,9 +12,9 @@ The example local environment configurations outlined below are based on
 [Lando installation instructions](https://docs.lando.dev/getting-started/installation.html).
 
 Depending on your role, you may only need a local backend environment, or you
-may instead prefer a local environment that can run both the back-end and
-front-end. The following sections outline the steps to set up a local
-environment for each of these use cases.
+may instead prefer a local environment that can run both the backend and
+frontend. The following sections outline the steps to set up a local environment
```
CobyPear, line 16:
```suggestion
front end. The following sections outline the steps to set up a local environment
```
Here it would not be hyphenated since it's not an adjective.

CobyPear (changes requested):
From the "Coyier tweet" image, my interpretation is that `frontend` should be `front end` in some cases and `front-end` in others, but not `frontend`.
So I think "front-end starter" or "front-end sites" is fine and when not a adjective "front end" is fine. I agree on leaving the URLs as is.

backlineint:
> From the "Coyier tweet",  my interpretation is that `frontend` should be `front end` in some cases and `front-end` in others, but not `frontend`. So I think "front-end starter" or "front-end sites" is fine and when not a adjective "front end" is fine. I agree on leaving the URLs as is.

@CobyPear Probably could have made the rationale here a little clearer. Primarily I'm aiming for consistency  - not just in the context of these docs, but also with the platform docs. My understanding is that the platform docs standardized on Google's guidance here: https://developers.google.com/style/word-list#frontend (they seem to lead away from the use of hyphens in general - https://developers.google.com/style/word-list#closed-compounds-and-prefixes.) 

Given how much of these docs have been re-purposed for the platform docs, I think consistency across the two sets of docs is worthwhile. I wouldn't be shocked if the platform docs change their stance at some point, but I think they should lead the way here.

Another alternative is that rather than changing now, we do nothing. This has already been more of a rabbit hole than I expected.

CobyPear:
> > From the "Coyier tweet",  my interpretation is that `frontend` should be `front end` in some cases and `front-end` in others, but not `frontend`. So I think "front-end starter" or "front-end sites" is fine and when not a adjective "front end" is fine. I agree on leaving the URLs as is.
> 
> @CobyPear Probably could have made the rationale here a little clearer. Primarily I'm aiming for consistency - not just in the context of these docs, but also with the platform docs. My understanding is that the platform docs standardized on Google's guidance here: https://developers.google.com/style/word-list#frontend (they seem to lead away from the use of hyphens in general - https://developers.google.com/style/word-list#closed-compounds-and-prefixes.)
> 
> Given how much of these docs have been re-purposed for the platform docs, I think consistency across the two sets of docs is worthwhile. I wouldn't be shocked if the platform docs change their stance at some point, but I think they should lead the way here.
> 
> Another alternative is that rather than changing now, we do nothing. This has already been more of a rabbit hole than I expected.

Okay that makes sense. I was not sure what style guide was landed on re: hyphens or not.

CobyPear (approved):

PR #909, opened 2023-10-20
KIT-46: `react-kit` storybook

CobyPear:
## What changes were made?
- Initialized storybook in `react-kit`
- Added some more scripts to the root package.json for react-kit
- Remove barrel file export (subpath exports only, can revert this when we start publishing if needed)
- Change css filename otherwise tsup builds 2 copies of it
- Added react plugin to the vite config to fix complaining tests

`react-kit`, `canary-sites-split` workflow.

## Additional information
Should we create a different workflow for these sites, or continue sticking them in the `canary-sites-split` workflow?

For some reason the typescript path aliases don't work in the stores in a way that satisfies ESLint. When importing a component it is typed as "any" and eslint complains even though it works during dev and build mode. Easy workaround is to use the relative path. This is fine but I really just want these to work all of the time :'(


CobyPear:
Using a branch on the repo instead of from the fork.

PR #927, opened 2023-11-03
KIT-92: [react-kit] Header component

CobyPear:
## What changes were made?
- Added a Header component, and the subcomponents required for it like the Button, IconButton, and Row.
- Fix eslint workspace settings
- Fix typescript path aliases for react-kit
- Add button component and stories
- Add svgs and global type file for storybook stories
- Upgrade workspace vite version to 4.5
- Use @vitejs/plugin-react-swc instead of @vitejs/plugin-react

react-kit
Storybook & tests
## Additional information

We might want to add another workspace or some way to share assets between packages/workspaces like the logo.


packages/react-kit/src/stories/Header.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+	args: {
+		Logo,
+		children: <NavItems />,
+	},
+	argTypes: {
+		children: {
+			control: {
+				type: 'radio',
+			},
+			options: ['Default', 'Many'],
```
backlineint, line 132: Including the 'many' example here is extremely useful. I was expecting this to wrap differently

nav_wrap

We'll probably have to talk through this as a group since this is a little hard to describe in the PR. But I was expecting the links and the buttons to be laid out as individual flexbox elements. The links would then wrap within their container, but not expand into the area containing the buttons. All elements would vertically center as the wrapped links expand the height of the header.

This will also be easier if we have two nav elements.

packages/react-kit/src/stories/Header.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const Logo = () => {
+	return (
+		<a href="/">
+			<img
+				src={PantheonLogo}
+				width="40"
+				height="40"
+				alt="Pantheon Systems"
+				className="lg:rk-mr-8"
```
backlineint, line 17: We'll have to talk through this with Cindy, but seeing this in browser I'm not convinced the nudge of margin the design gives to the logo and the icon buttons works. It makes things look misaligned to me. I'd expect to have a similar feeling when we start to see this in page layouts.

logo_alignment

backlineint: I also wonder about how much of things like pixel widths and tailwind alignment should be the responsibility of the consumer here. It's a tradeoff though - the more we own, the more likely they will get it right, but it will also result in less control for the user.

packages/react-kit/src/stories/Header.stories.tsx
backlineint, line 1: On extremely wide viewports the header is left aligned.

wide_view

In this context a little hard to say if this is a problem, but it is not what I would expect. If this header is meant to work 'by itself' I'd expect it to be centered (this existence of the row component implies this to me.) If the header needs to be included in a layout for this to work, I'd expect the component in storybook to always be full width with the gap between the two sets of links continuing to expand. In that case, I'd have some questions about the use of the row - I'd expect the margins to be part of the layout rather than the row inside the header component.

backlineint (changes requested):
A lot of great stuff here. The storybook implementation is invaluable for reviewing this. The attention to a11y details is also excellent.

Also a decent amount of visual issues for us to discuss and refine :)

The biggest fundamental thing at the root of a lot of this is that I think this component should have two distinct nav elements - the links (primary nav) and the buttons (secondary nav). I think that change will make it easier to address many of the other issues I mentioned inline.

packages/react-kit/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+	/**
+	 * Need to use a ref here so the FocusTrap can work properly.
+	 */
+	const NavList = React.forwardRef<
+		React.ElementRef<'nav'>,
+		{ children: React.ReactNode }
+	>(({ children }, ref) => {
+		return (
+			<nav
```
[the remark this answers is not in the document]
CobyPear: That makes sense. I was struggling with that part of the design. Splitting it up would solve it nicely.

packages/react-kit/src/stories/Header.stories.tsx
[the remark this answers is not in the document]
CobyPear: Sounds good to me, will do.

packages/react-kit/src/components/Row/index.tsx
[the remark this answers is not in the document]
CobyPear: Yeah good call, I have some stories from the poc I can reuse here at the very least for now.

packages/react-kit/src/stories/Header.stories.tsx
CobyPear: > If this header is meant to work 'by itself' I'd expect it to be centered (this existence of the row component implies this to me.)

This might be a config thing on the story itself. It can be centered if that's easier to reason about.

> with the gap between the two sets of links continuing to expand

I may be misunderstanding what these stories should be accomplishing. Am I to be showing the component as the design is, or just the component in the most plain way it can exist? The design has a max width of 1920px so I don't agree that the links should keep expanding and the nav should always be full width, but again that depends on the goal of these stories.

packages/react-kit/src/stories/Header.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+			<hr
+				className="divide-x rk-my-8 rk-flex rk-w-full lg:rk-hidden"
+				key="hr"
+			/>
+		</li>,
+		<li
+			className="rk-mb-3 rk-mr-auto lg:rk-mb-0 lg:rk-ml-auto lg:rk-mr-0"
+			key="docs"
+		>
+			<Button
```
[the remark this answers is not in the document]
CobyPear: It's possible the small buttons aren't the right size, but these are the `small` buttons.

packages/react-kit/src/components/Header/Header.tsx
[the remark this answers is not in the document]
CobyPear: I may need some more guidance here on the markup - are you envisioning something like this?

```tsx
<header>
	<Logo />
	<nav aria-label="Main">
		<ul>{mainNavItems}</ul>
	</nav>
	<nav aria-label="Secondary">
		<ul>{secondaryNavItems}</ul>
	</nav>
	...buttons
</header>;


```

packages/react-kit/src/stories/Header.stories.tsx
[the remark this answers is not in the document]
CobyPear: Buttons were wrong, fixed.

packages/react-kit/src/stories/Header.stories.tsx
CobyPear: `mx-auto` saves another day

CobyPear:
@backlineint I attempted to generate the typedocs for these changes, but the next upgrade seems to have broken that yet again :)

CobyPear:
Was able to work around the nextjs kit docs errors and generated the api reference for `react-kit` only.

packages/react-kit/src/components/Header/Header.tsx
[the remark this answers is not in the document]
backlineint: The buttons would be in the second nav, but otherwise this is what I was expecting. The actual markup I'm seeing in the latest branch looks right to me.

packages/react-kit/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,145 @@
...
+						isOpen && windowWidth <= MAX_MOBILE_NAV_WIDTH
+							? 'rk-absolute rk-bottom-0 rk-left-0 rk-z-10 rk-h-[calc(100%_-_72px)] rk-flex-col rk-overflow-y-auto rk-overflow-x-hidden rk-px-4 rk-pb-4 rk-pt-8 sm:rk-px-6 lg:rk-px-12'
+							: 'rk-hidden rk-flex-wrap lg:rk-flex',
+					)}
+				>
+					<nav
+						className={clsx(NAVLIST_STYLES.nav, 'rk-mr-auto lg:rk-w-2/3')}
+						aria-label="Main"
+					>
+						<ul className={NAVLIST_STYLES.ul}>{mainNavItems}</ul>
```
backlineint, line 85: I still think that requiring the consumer to assemble the markup for the entire main and secondary nav lists (including some pretty complicated tailwind classes) pushes too much responsibility outside of this component. Rather than full markup, I'd expect the prop to be an array of nav items similar to the `navItems` list used in the storybook example. And then the component takes care of juggling all of the tailwind classes automatically.

It is a tradeoff though - more likely that they will get the expected end result, but less control in overriding look and feel. In a non-React project I'd use `slot` elements as an escape hatch here. If a name slot for either of the nav lists is provided, render that instead. I see that Radix has a slot concept, but based on the docs it isn't clear to me if it can support multiple named slots like we'd need here.

packages/react-kit/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+		return (
+			<header
+				ref={ref}
+				className={clsx(
+					navbarStyles,
+					'rk-z-10 rk-flex rk-w-full rk-items-center rk-py-4',
+				)}
+			>
+				<Logo />
```
backlineint, line 71: This is a smaller (and thus lower priority) case of putting too much markup responsibility on the consumer as discussed below in the comment about the nav items. If we come up with a way to solve that using slots, I wonder if that same approach could apply here. If not, that is probably OK - markup here is less complicated, and less likely to change.

packages/react-kit/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,145 @@
...
+					className={clsx(
+						overlayStyles,
+						'rk-w-full rk-justify-around',
+						isOpen && windowWidth <= MAX_MOBILE_NAV_WIDTH
+							? 'rk-absolute rk-bottom-0 rk-left-0 rk-z-10 rk-h-[calc(100%_-_72px)] rk-flex-col rk-overflow-y-auto rk-overflow-x-hidden rk-px-4 rk-pb-4 rk-pt-8 sm:rk-px-6 lg:rk-px-12'
+							: 'rk-hidden rk-flex-wrap lg:rk-flex',
+					)}
+				>
+					<nav
+						className={clsx(NAVLIST_STYLES.nav, 'rk-mr-auto lg:rk-w-2/3')}
```
backlineint, line 82: Big improvement here, but 2/3 / 1/3 seems to give a little too much space between the nav elements.

flex-between

Could we get away with something like 3/4 / 1/4?

backlineint (changes requested):

packages/react-kit/src/components/Header/Header.tsx
CobyPear: Are you referring to the `Logo` component here? GitHub is making it hard to tell.

packages/react-kit/src/components/Header/Header.tsx
CobyPear: Implemented something that should allow a tuple of labels and hrefs or markup.

packages/react-kit/src/components/Header/Header.tsx
backlineint: Yes - was referring to the logo here.

packages/react-kit/src/assets/pantheon-fist-blk.svg
backlineint, line 1: Might have gotten our wires crossed here. I'd expect the entire image to be tightly cropped and any margin or spacing to be handled in css.

packages/react-kit/src/components/Header/Header.tsx
```diff
@@ -58,6 +58,53 @@ export const Header = ({
...
+			overlay: clsx(
+				'rk-w-full',
+				isOpen && windowWidth <= MAX_MOBILE_NAV_WIDTH
+					? 'rk-absolute rk-bottom-0 rk-left-0 rk-z-10 rk-h-[calc(100%_-_72px)] rk-flex-col rk-overflow-y-auto rk-overflow-x-hidden rk-px-4 rk-pb-4 rk-pt-8 sm:rk-px-6 lg:rk-px-12'
+					: 'rk-hidden rk-flex-wrap lg:rk-flex',
+			),
+		};
+
+		const PrimaryNav = () => {
+			if (isNavTuple(mainNavItems)) {
```
backlineint, line 70: This does a nice job of allowing both the 'just works' and 'full control' use cases. Likely to be an approach we can use elsewhere. Thanks!

backlineint (changes requested):
I think we're down to two things here:
* Getting a correctly cropped logo asset and then styling that appropriately.
* Fixing the alignment of the `hr` in the expanded mobile menu.

backlineint (approved):

PR #954, opened 2023-11-21
KIT-111: Prevent Caching In Test Preview Route

mitchellmarkoff:
## What changes were made?
- Prevents refresh and clears the `access_token` if `test=true` in the preview route.

`next-drupal-starter`

Locally
## Additional information

### DIFF
--- a/packages/create-pantheon-decoupled-kit/src/templates/next-drupal/pages/api/preview.js
+++ b/packages/create-pantheon-decoupled-kit/src/templates/next-drupal/pages/api/preview.js
@@ -30,7 +30,7 @@ const preview = async (req, res) => {
 		let content;
 		try {
 			//  Clear access token
-			delete store.token.accessToken;
+			Object.keys(store.token).forEach((v) => (store.token[v] = ''));
 			content = await store.getObjectByPath({
 				objectName,
 				path: slug,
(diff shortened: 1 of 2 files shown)


packages/create-pantheon-decoupled-kit/src/templates/next-drupal/pages/api/preview.js
```diff
@@ -44,6 +45,10 @@ const preview = async (req, res) => {
 				message: error.message,
 			});
 		}
+
+		//  Clear access token
+		delete store.token.accessToken;
```
CobyPear, line 50: I think we'll want to do this before the call to getObjectByPath, because if that fails this line won't get called and the subsequent call may have the cached token.

CobyPear (changes requested):

mitchellmarkoff:
> This seems to work fine on the first request, but if I refresh the page it fails with a 401. For some reason it seems like the store is not re-fetching the access token. I wonder if we need to fix this upstream or if we can figure out a workaround here.

It looks like this is because DrupalState generates a new token based off of the `validUntil` field so since the `accessToken` was the only thing getting cleared, DrupalState thought there was a valid token and didn't regenerate a new one. This new commit should solve that issue. The token fields need to be set to `''` rather than deleted since DrupalState checks those and will error if they are undefined.

CobyPear (changes requested):
LGTM, just needs a changeset (I almost forgot too!)

CobyPear (approved):
The changeset names never get old 🦎 🔨

PR #955, opened 2023-11-21
KIT-72: Implement new header component in the `next-wordpress` starter

CobyPear:
## What changes were made?
- Re-export `react-kit` components through `nextjs-kit`
- Fix typescript module resolution for some packages - we should be using bundler in almost all of our cases
- Fix package.json where needed
- Resolve `wordpress-kit` unhandled error if headers are undefined - could be more robust but works for now
- Implement the new header component in the Layout for `next-wp-starter`

`nextjs-kit`, `react-kit`, `wordpress-kit`, `cli`

locally

## Additional information

For the `href` of the  header links, WordPress currently does not offer any additional metadata about the menu item that could be used on the frontend to determine if it is a post or a page, so right now the link is broken (as it has been in the footer since we implemented this). We could hardcode the prefix but ideally it would come from the CMS like we do with Drupal. Should we fix this on the frontend in the short term or wait for the fix in the CMS?


backlineint:
> I was sort of expecting to need to re-implement some parts of the header or buttons where a Next.js <Link> component may be necessary. Thinks seem to work with just a normal <a> tag but are we missing some functionality without using <Link>? We can always use asChild. I figure we will refine this as we implement this across the templates, and eventually we'll need to clean up the CSS modules which will be another opportunity to answer these questions.

The main thing we lose by not using the `Link` component is pre-fetching of pages, which would be nice to have. My understanding is that we are using `a` instead of `Link` currently, but the LinkComponent prop would make it possible to use `Link` instead.  If I'm correct about that, I'd say we should make that change now.

CobyPear:
> > I was sort of expecting to need to re-implement some parts of the header or buttons where a Next.js  component may be necessary. Thinks seem to work with just a normal  tag but are we missing some functionality without using ? We can always use asChild. I figure we will refine this as we implement this across the templates, and eventually we'll need to clean up the CSS modules which will be another opportunity to answer these questions.
> 
> The main thing we lose by not using the `Link` component is pre-fetching of pages, which would be nice to have. My understanding is that we are using `a` instead of `Link` currently, but the LinkComponent prop would make it possible to use `Link` instead. If I'm correct about that, I'd say we should make that change now.

I forgot to update the PR but I did end up implementing the Link component into the Header exported from `nextjs-kit`

packages/create-pantheon-decoupled-kit/src/templates/next-wp/lib/Menus.js
```diff
@@ -29,8 +29,44 @@ export async function getFooterMenu() {
...
+			headers,
+		} = await client.rawRequest(query);
+
+		if (!menu) {
+			throw new Error(
+				'No header menu data available. Try customizing your query.',
+			);
+		}
+		const headerMenuItems = menu?.menuItems?.edges?.map(({ node }) => node);
+		return { headerMenuItems, headerMenuItemHeaderes: headers };
```
backlineint, line 69: @CobyPear happened to notice this after merging. `headerMenuItemHeaderes` - Headeres looks like a typo.

PR #962, opened 2023-11-28
KIT-63: Adjust Gatsby Health Check Environment Variable Valid Protocol

mitchellmarkoff:
## What changes were made?
Health check will not fail if CMS is http but will fail if no protocol is set.

`decoupled-kit-health-check`

locally

## Additional information

Test steps can be found in the comments of the related ticket.

### DIFF
--- a/packages/decoupled-kit-health-check/__tests__/GatsbyWordPressHealthCheck.test.ts
+++ b/packages/decoupled-kit-health-check/__tests__/GatsbyWordPressHealthCheck.test.ts
@@ -101,4 +101,18 @@ describe('GatsbyWordPressHealthCheck', () => {
 
 		expect(url.href).toEqual('https://wordpress.test/wp/graphql');
 	});
+	it('should fail if no protocol is set for the WPGRAPHQL_URL', async ({
+		logSpy,
+	}) => {
+		process.env['WPGRAPHQL_URL'] = 'wordpress.test';
+		const HC = new GatsbyWordPressHealthCheck({ env: process.env });
+		await HC.validateEndpoint()
+			.then((hc) => hc.validateWPGatsbyPlugin())
+			.then((hc) => hc.validateMenu())
+			.then((hc) => hc.validateAuth())
+			.catch(console.error);
+
+		const result = logSpy.mock.calls.map(([call]) => call);
+		expect(result).toMatchSnapshot();
+	});
 });
--- a/packages/decoupled-kit-health-check/__tests__/GatsbyWordPressHealthCheck.test.ts
+++ b/packages/decoupled-kit-health-check/__tests__/GatsbyWordPressHealthCheck.test.ts
@@ -110,7 +110,7 @@ describe('GatsbyWordPressHealthCheck', () => {
 			.then((hc) => hc.validateWPGatsbyPlugin())
 			.then((hc) => hc.validateMenu())
 			.then((hc) => hc.validateAuth())
-			.catch(console.error);
+			.catch((err) => console.log(err.message));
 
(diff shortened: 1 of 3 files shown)


packages/decoupled-kit-health-check/__tests__/GatsbyWordPressHealthCheck.test.ts
```diff
@@ -27,7 +27,7 @@ describe('GatsbyWordPressHealthCheck', () => {
 	it('should pass for a valid PANTHEON_CMS_ENDPOINT and invalid auth', async ({
 		logSpy,
 	}) => {
-		process.env['PANTHEON_CMS_ENDPOINT'] = 'wordpress.test';
+		process.env['PANTHEON_CMS_ENDPOINT'] = 'https://wordpress.test';
```
CobyPear, line 30: the `PANTHEON_CMS_ENDPOINT` doesn't have the protocol attached in prod which is what is being modeled here. It's important for the health-check to work in that case and add the protocol if needed since this is the default state when spinning up a new BE + FE through the dashboard.

packages/decoupled-kit-health-check/src/classes/GatsbyWordPressHealthCheck.ts
```diff
@@ -73,9 +73,7 @@ export class GatsbyWordPressHealthCheck extends WordPressHealthCheck {
 			? Object.entries(backendVars).filter(([key]) => key === 'WPGRAPHQL_URL')
 			: Object.entries(backendVars);
 
-		const url = /^https:\/\//.test(endpoint) ? endpoint : `https://${endpoint}`;
-
-		this.endpoint = url;
+		this.endpoint = endpoint;
```
CobyPear, line 76: I think we still need some logic to determine the protocol here, just need to account for `http` as well:
```suggestion
		const url = /^https?:\/\//.test(endpoint) ? endpoint : `https://${endpoint}`;
```

CobyPear (changes requested):

packages/decoupled-kit-health-check/src/classes/GatsbyWordPressHealthCheck.ts
mitchellmarkoff: @CobyPear  Ahh I now see why this is necessary considering the `PANTHEON_CMS_ENDPOINT` endpoint in prod. However this line is what is allowing the health check to pass if no protocol is set for the `WPGRAPHQL_URL`. Perhaps a solution would be to only use this logic on the `PANTHEON_CMS_ENDPOINT`.

packages/decoupled-kit-health-check/__tests__/__snapshots__/GatsbyWordPressHealthCheck.test.ts.snap
```diff
@@ -1,5 +1,14 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`GatsbyWordPressHealthCheck > should fail if no protocol is set for the WPGRAPHQL_URL 1`] = `
+[
+  "No .env* file found, assuming production environment.",
+  "Checking for PANTHEON_CMS_ENDPOINT or WPGRAPHQL_URL...",
+  "⎩✅ WPGRAPHQL_URL is set!",
+  "Validating CMS endpoint...",
```
CobyPear, line 8: We should throw an error in this case, probably `InvalidCMSEndpointError`

CobyPear (changes requested):

CobyPear (approved):
LGTM

PR #963, opened 2023-11-28
KIT-41: [next-drupal] Implement Header

CobyPear:
## What changes were made?
Implement new Header in the next-drupal-starter and the next-drupal add-ons. Search was not tested as there will be extra consideration for the header and the new search component.

When we implement the footer, we can remove the footer menu items from the layout, in the `next-drupal` and `next-wp` templates.
Locally using the monorepo, will check again with the snapshot release version of the CLI when it is up.

## Additional information


backlineint (approved):
Created one styling follow up ticket that we can consider: https://getpantheon.atlassian.net/browse/KIT-136

PR #977, opened 2023-12-05
KIT-93: [react-kit] `<Footer />` component

mitchellmarkoff:
## What changes were made?
Adds an implementation of a `<Footer />` component to `react-kit` per the re-theme designs.

`react-kit`

Storybook & tests

## Additional information


packages/react-kit/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,37 @@
...
+		if (React.isValidElement(Logo)) {
+			return Logo;
+		} else if (typeof Logo === 'object' && 'src' in Logo && 'alt' in Logo) {
+			const { src, alt } = Logo;
+			return <img src={src} height="40" alt={alt} className="rk-inline" />;
+		}
+		return null;
+	};
+	return (
+		<footer className="rk-mt-16 rk-bg-neutral-800 rk-py-4 rk-text-center rk-text-white">
```
CobyPear, line 18: There should be no top margin here

packages/react-kit/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { FooterProps } from './props';
+
+/**
+ * @see {@link https://live-storybook-react-kit.appa.pantheon.site/?path=/docs/footer-footer--docs}
+ */
+export const Footer = ({ Logo, Link }: FooterProps) => {
```
CobyPear, line 7: I would add a prop for `className` so styles could be passed in to the `<footer>`

packages/react-kit/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,37 @@
...
+		}
+		return null;
+	};
+	return (
+		<footer className="rk-mt-16 rk-bg-neutral-800 rk-py-4 rk-text-center rk-text-white">
+			<div className="rk-mb-0.875 rk-mx-auto rk-mt-2 rk-block">
+				<FooterLogo />
+			</div>
+
+			<h6 className="rk-mb-0.875 rk-text-lg rk-font-bold">
```
CobyPear, line 23: Probably don't need a heading tag here.

packages/react-kit/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,37 @@
...
+			</h6>
+
+			<p className="rk-mb-0.875 rk-text-sm">© Pantheon 2013</p>
+
+			<p className="rk-mb-2 rk-text-xs">
+				Built with{' '}
+				<a href={Link} className={'rk-underline'}>
+					Pantheon.io
+				</a>
+			</p>
```
CobyPear, line 34: These should probably be spans/divs since there is no semantic paragraph being represented.

packages/react-kit/src/components/Footer/props.ts
```diff
@@ -0,0 +1,24 @@
...
+	 *
+	 */
+	Logo:
+		| { src: string; alt: string; href: string; styles?: string }
+		| React.ReactElement
+		| JSX.Element;
+	/**
+	 * Link used for company website.
+	 */
+	Link: string;
```
CobyPear, line 23: Small thing but this should probs be lowercase. I wonder if we should make this more generic in case someone doesn't want to put a link here, but we can save that for a future enhancement.

CobyPear (changes requested):
image
image


The horizontal space seems off in a few places - looks like the elements should make it 300px tall, currently it's at 156px.

We'll need a white version of the logo. id if it's as easy as copying the current svg and changing the fill colors or if it's more involved, you can export the svg from the Figma file.

The text sizes seem off compared to the design.

packages/react-kit/src/components/Footer/Footer.tsx
CobyPear: Hmm true but should we implement it on the footer itself or on the page/pages that the footer resides?

mitchellmarkoff:
Lots of updated to the structure of the component here. I played around with the styles and this should be the correct sizing now (landed at 299px tall after confirming the correct spacing was being used), the text still looks slightly different compared to the designs but I am using the specified text sizes outlined in Figma so I am not sure what that is attributed to exactly.

packages/react-kit/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+			);
+		}
+
+		return null;
+	};
+	return (
+		<footer
+			className={clsx(
+				className,
+				'rk-bg-neutral-800 rk-pt-3.5 rk-text-center rk-text-white',
```
CobyPear, line 66: It's unclear to me why this top padding is here

packages/react-kit/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+			'builtWithLink' in Content
+		) {
+			return (
+				<>
+					<div className="rk-text-base rk-font-bold">{Content.title}</div>
+
+					<div className="rk-pb-8 rk-text-base">{Content.copy}</div>
+
+					<div className="rk-pb-14 rk-text-sm">
+						Built with{' '}
```
CobyPear, line 48: I'm not sure we even want to hardcode the 'Built with' text – that should all come in from props. I'd expect a blank slate in this section to work with here if using the component from scratch. This is making me go back to thinking we should accept `children` instead of a component as a prop.

packages/react-kit/src/components/Footer/props.ts
```diff
@@ -0,0 +1,27 @@
...
+		| React.ReactElement
+		| JSX.Element;
+	/**
+	 * Styles to be passed to the footer component.
+	 */
+	className?: string;
+	/**
+	 *  An instance of React.ReactChild
+	 */
+	children: JSX.Element;
```
CobyPear, line 26:
```suggestion
	children: React.ReactNode;
```
`ReactNode` should be used for the `children` type

CobyPear (changes requested):
One small thing and I think this is good to go after that. Looks great!

CobyPear (approved):

PR #984, opened 2023-12-11
KIT-142: Fix gatsby develop infinite re-render

CobyPear:
## What changes were made?
- Fix the tailwindcss config for gatsby so that it does not force an infinite re-render in dev mode due to the `d.ts` files
`cli` templates.

## Additional information


mitchellmarkoff (approved):
LGTM 🏄🏼‍♂️

PR #985, opened 2023-12-11
KIT-173: Use GHA instead of CircleCI for test job

CobyPear:
## What changes were made?
First step towards removing CircleCI and adopting Dependabot for this repo.
We may need to adjust the branch rules regarding the CircleCI check.

- Remove test job from circleci config
- Add new GitHub Action workflow to run the tests on push to any branch
  - Do we want to lock this down to PRs or certain branches?

## Additional information

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,10 +97,3 @@ workflows:
     jobs:
       - audit:
           context: decoupled-slack-secrets
-  orb-free-workflow:
-    when:
-      not: # Run test job when trigger is not the scheduled pipeline
-        equal: [scheduled_pipeline, << pipeline.trigger_source >>]
-    jobs:
-      - test:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
backlineint, line 1: In the short term before we remove this entirely, we probably need to remove the `orb-free-workflow` workflow to prevent circle from complaining.

backlineint (changes requested):
Just needs one small change so Circle stops yelling at us.

backlineint:
> Do we want to lock this down to PRs or certain branches?

Can't think of any reason we need to. Can always restrict this if it becomes too noisy.

.circleci/config.yml
CobyPear: ah yes, I knew I was missing something!

backlineint (approved):

PR #1108, opened 2024-04-15
Remove deps from test fixtures

CobyPear:
## What changes were made?
Removed the deps from the test fixtures in the cli package since there is no way to ignore certain subdirs in the dependabot config.

These were causing false positives for dependabot. They aren't necessary for our tests.
Changed the vitest coverage.all option to false

Also made the reporter only report on tested files.

`cli`

## Additional information


mitchellmarkoff (approved):
Nice cleanup work 🌮

